PR #2, opened 2019-08-22
Add deploy script

JosephusPaye:
This adds a script that builds the site and deploys to Github Pages.

To deploy, run:

```bash
yarn deploy
```

TODO:

- [x] Prevent deploys from non-master branches


JosephusPaye:
> Looks good. Although one thing, I'm wondering if we should have more comments on things? Document the documentation :P

Yeah true, which parts are unclear and need comments?

PR #15, opened 2019-09-18
Changes .gitattributes to be more friendly for web projects

ysims:
This pull request fixes issues caused by the '.gitattributes' file. 

Files that are not text files (binary) will no longer have characters assumed to be CR-LF (0d0a) turned into LF-only (0a).


.gitattributes
```diff
@@ -1 +1,27 @@
-* text=auto eol=lf
\ No newline at end of file
+## GITATTRIBUTES FOR WEB PROJECTS
```
JosephusPaye: This header comment is unnecessary IMO.

.gitattributes
```diff
@@ -1 +1,27 @@
...
+#   binary  These files are binary and should be left untouched.
+#
+# Note that binary is a macro for -text -diff.
+######################################################################
+
+# Auto detect
+##   Handle line endings automatically for files detected as
+##   text and leave all files detected as binary untouched.
+##   This will handle all files NOT defined below.
+*                 text=auto
```
JosephusPaye: I think we're missing the `eol=lf` option

.gitattributes
ysims: Fixed up the comments in latest commit

.gitattributes
ysims: text=auto does this

JosephusPaye: True 😅

PR #17, opened 2019-09-26
Auto scroll sidebar to active link on load

JosephusPaye:
This PR ensures that the active link in the sidebar is visible by scrolling to it on page load.


JosephusPaye:
@ysims I'm tempted to abandon this - most solutions I can think of have caveats that make them not always work. May not be worth the effort and additional code.

BrendanAnnable:
Could you use [Element#ScrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView)?

JosephusPaye:
Was just about to reach out, Brendan.

I think I tried `Element#ScrollIntoView` and got mixed results but don't recall what the problem was exactly.

I'll give it another go.

JosephusPaye:
I think the issue I was concerned about was the scrollbar visibly jumping from the top when navigating pages:

nubook-sidebar-scroll

Seems like something we could live with though.

JosephusPaye:
@ysims Could you try to see if it has the same problem you saw before?

ysims:
The problem has gone now :)

PR #35, opened 2019-12-08
Adds NUClear page

ysims:
Adds a page which gives an overview of various aspects of the NUClear software framework.

- [x] Emitting section
- [x] Managing reactions section
- [ ] Examples section

[Preview.](https://deploy-preview-35--nubook.netlify.com/system/nubots/nuclear)


src/book/02-system/01-nubots/03-nuclear.mdx
```diff
@@ -5,3 +5,142 @@ title: NUClear
...
+});
+```
+
+In the above request, the Trigger on dataType B has been listed as optional, while the Trigger for dataType A is listed as mandatory. Yet the callback function lists arguments for both dataType A and dataType B.
+
+Lets say that dataType A is emitted to the PowerPlant, but at this time, dataType B does not have any data associated with it.
+
+Since dataType B was listed as optional, the task associated with this reaction can be scheduled. However, when executing the callback for this reaction, NUClear will identify that dataType B is not present, and will remove reference to this data type from the callback, so that the task is only run for dataType A.
+
+Effectively, through the application of fission, the callback is restructured as per the following example.
```
JosephusPaye:
```suggestion
Effectively, through the application of fission, the callback is restructured to look like the following example.
```

src/book/02-system/01-nubots/03-nuclear.mdx
```diff
@@ -5,3 +5,142 @@ title: NUClear
...
+});
+```
+
+## Emit Statements
+
+Emit statements are used by Reactors wanting to emit data to the PowerPlant. When using NUClear, data will most likely be emitted during a reaction. However, where necessary, emissions can also occur during reactor construction, or in some cases from within the PowerPlant itself.
+
+Any data emitted to the PowerPlant will be sent with a unique pointer. The PowerPlant will take ownership of this pointer and run any necessary callbacks to trigger reactions.
+
+Data can be emitted under varying scopes. These can be local or network emissions. More can be found on emit statements on the [NUClear readthedocs page](https://nuclear.readthedocs.io/en/latest/dsl.html#emit-statements).
```
JosephusPaye:
```suggestion
Data can be emitted under varying scopes. These can be local or network emissions. More can be found on emit statements in the [NUClear documentation](https://nuclear.readthedocs.io/en/latest/dsl.html#emit-statements).
```

src/book/02-system/01-nubots/03-nuclear.mdx
```diff
@@ -5,3 +5,142 @@ title: NUClear
...
+
+### Local Emitting
+
+These emissions send data to the local instance of the NUClear PowerPlant. There are a number of scopes under which these emissions can take place. The syntax for these emissions are `emit<Scope::SCOPE>(data, dataType);`, where `SCOPE` is replaced with a scope as defined below.
```
JosephusPaye:
```suggestion
Local emissions send data to the local instance of the NUClear PowerPlant. Essentially this is the current running process. There are a number of scopes under which these emissions can take place. The syntax for these emissions are `emit<Scope::SCOPE>(data, dataType);`, where `SCOPE` is replaced with a scope as defined below.
```

src/book/02-system/01-nubots/03-nuclear.mdx
```diff
@@ -5,3 +5,142 @@ title: NUClear
...
+
+## Emit Statements
+
+Emit statements are used by Reactors wanting to emit data to the PowerPlant. When using NUClear, data will most likely be emitted during a reaction. However, where necessary, emissions can also occur during reactor construction, or in some cases from within the PowerPlant itself.
+
+Any data emitted to the PowerPlant will be sent with a unique pointer. The PowerPlant will take ownership of this pointer and run any necessary callbacks to trigger reactions.
+
+Data can be emitted under varying scopes. These can be local or network emissions. More can be found on emit statements on the [NUClear readthedocs page](https://nuclear.readthedocs.io/en/latest/dsl.html#emit-statements).
+
+### Local Emitting
```
JosephusPaye:
```suggestion
### Local Emissions
```

src/book/02-system/01-nubots/03-nuclear.mdx
```diff
@@ -5,3 +5,142 @@ title: NUClear
...
+| ------- | ------------------------------------------------------------------------------------------------------------------------------------ |
+| UDP     | Emits data as a UDP packet over the network. Is of the form `emit<Scope::UDP>(data, to_addr, to_port);`.                             |
+| NETWORK | Emits data over the network to other NUClear environments. Is of the form `emit<Scope::NETWORK>(data, target, reliable, dataType);`. |
+
+# NUClear Roles
+
+The NUClear roles system is a build and messaging system for the NUClear framework. It uses [CMake](https://cmake.org/) and [Python](https://www.python.org/) to manage the construction of various executables made up of a selection of modules. These executables are called roles.
+
+CMake is used as the main system for generating the libraries and executables that are used for the final binary. Note that it utilises [globbing](<https://en.wikipedia.org/wiki/Glob_(programming)>) to find the sources that are used for modules. If you add or remove a file, you must rerun cmake so that it locates the new files.
```
JosephusPaye:
```suggestion
CMake is used as the main system for generating the libraries and executables that are used for the final binary. Note that it utilises [globbing](<https://en.wikipedia.org/wiki/Glob_(programming)>) to find the sources that are used for modules. If you add or remove a file, you must rerun CMake to locate the new files.
```

src/book/02-system/01-nubots/03-nuclear.mdx
```diff
@@ -5,3 +5,142 @@ title: NUClear
...
+
+# Managing Reactions
+
+During system runtime, executing reactions can be managed via the reaction handle. A reaction handle is provided for any binding On Statements. Once an on statement has been bound, the reaction handle will be enabled. If necessary, reactions can toggle between enabled and disabled during runtime.
```
JosephusPaye:
```suggestion
During system runtime, executing reactions can be managed via their associated reaction handles. A reaction handle is provided for binding on statements. Once an on statement has been bound, the reaction handle will be enabled. If necessary, reactions can toggle between enabled and disabled during runtime.
```

src/book/02-system/01-nubots/03-nuclear.mdx
```diff
@@ -5,3 +5,142 @@ title: NUClear
...
+
+| Handle           | Description                                                                                                                                                                                         |
+| ---------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| disable()        | Disables the reaction. Tasks will not be created if triggered.                                                                                                                                      |
+| enable()         | Enables the reaction so that associated tasks will be scheduled and queued when the reaction is triggered.                                                                                          |
+| enable(bool set) | Sets the run status of the reaction handle. True will enable and false will disable the reaction.                                                                                                   |
```
JosephusPaye:
```suggestion
| enable(bool set) | Sets the run status of the reaction handle. `true` will enable the reaction and `false` will disable it.                                                                                                   |
```

src/book/02-system/01-nubots/03-nuclear.mdx
```diff
@@ -5,3 +5,142 @@ title: NUClear
...
+| enabled()        | Informs if the reaction is currently enabled. Returns true if enabled, false if disabled.                                                                                                           |
```
JosephusPaye:
```suggestion
| enabled()        | Determines the current status of the reaction. Returns `true` if the reaction is enabled, and `false` otherwise.                                                                                                           |
```

src/book/02-system/01-nubots/03-nuclear.mdx
```diff
@@ -5,3 +5,142 @@ title: NUClear
...
+| unbind()         | Removes a reaction request from the runtime environment. This action is not reversible, once a reaction has been unbound, it is no longer available for further use during the instance of runtime. |
```
JosephusPaye:
```suggestion
| unbind()         | Removes a reaction request from the runtime environment. This action is not reversible, once a reaction has been unbound, it is no longer available for further use during that runtime. |
```

src/book/02-system/01-nubots/images/power_plant.svg
```diff
@@ -0,0 +1,721 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
JosephusPaye: This SVG file can be optimized or a smaller size using [SVGOMG](https://jakearchibald.github.io/svgomg/).

Would also be nice, but not necessary, if it had a transparent background.

src/book/02-system/01-nubots/03-nuclear.mdx
```diff
@@ -5,3 +5,144 @@ title: NUClear
...
+| With     | Is used to define any extra data which should be provided to a subscribing a reaction. This should be fused with at least one other DSL word as the use of this word alone will not trigger a reaction within the system. `on<Trigger<T1>, With<T2>>()` will trigger when T1 is emitted into the system. It will give read-only access of T1 and T2 to the callback. If T2 is not present when T1 is emitted to the system, the reaction will not run. |
+| Last     | Instructs the PowerPlant to store the last $n$ messages (of the associated type) to the cache and provide read-only access to the subscribing reaction. `on<Last<n, Trigger<T, ...>>>()` will store the next $n$ messages of each of the `Trigger` types in a list with the oldest message first. After $n$ messages are stored, the trigger of a new reaction will cause the oldest message to be dropped and the newest copy appended to the list.                                         |
```
TrentHouliston: Probably a good idea to put here that Last caches the last N that were `Received` and not the last n that were `Sent`, so in the event that you're dropping data last won't have those dropped packets in it.

src/book/02-system/01-nubots/03-nuclear.mdx
```diff
@@ -5,3 +5,144 @@ title: NUClear
...
+| Optional | Is used to signify any optional properties in the DSL request. `on<Trigger<T1>, Optional<With<T2>()` will run when T1 is emitted into the system. If T2 is available, then the reaction will be given read-only access to the most recent emission of T2.                                                                                                                                                                                             |
```
TrentHouliston:
```suggestion
| Optional | Is used to signify any optional properties in the DSL request. `on<Trigger<T1>, Optional<With<T2>>>()` will run when T1 is emitted into the system. If T2 is available, then the reaction will be given read-only access to the most recent emission of T2.                                                                                                                                                                                             |
```

src/book/02-system/01-nubots/03-nuclear.mdx
```diff
@@ -5,3 +5,144 @@ title: NUClear
...
+| Single   | Is used to specify that only one instance of the associated reaction can execute at any given point during runtime. `on<Trigger<T, ...>, Single>()` will only allow one task for this reaction to be executed or in the queue at any given time. This should be fused with another DSL word, such as `Trigger` in the example.                                                                                                                                           |
```
TrentHouliston: Should mention that single is equivalent to Buffer<1>

src/book/02-system/01-nubots/03-nuclear.mdx
```diff
@@ -5,3 +5,144 @@ title: NUClear
...
+| Priority | Assigns a priority to any tasks created to determine scheduling order in the PowerPlant. `on<Trigger<T, ...>, Priority::HIGH>()` will schedule the task with `HIGH` priority when the reaction is triggered. Available priority settings are `REALTIME`, `HIGH`, `NORMAL`, `LOW`, and `IDLE`. This should be fused with another DSL word, such as `Trigger` in the example.                                                                                      |
+| Sync     | Sets the synchronisation for a group of tasks. `on<Trigger<T, ...>, Sync<Group>>()` will allow only one task from `Group` to execute at any given time. This is similar to the `Single` word, but applies to multiple reactions. This word helps prevent race conditions. This should be fused with another DSL word, such as `Trigger` in the example.                                                                                   |
```
TrentHouliston: It's not similar to the `Single` word, you can, in fact, use both of them together. `Single` means "if this reaction is running and more data comes in drop it" and `Sync` means if something in this group is already running then queue this task until it is finished. So if you combine together you can have a setup where another reaction in the group is running, a task for this reaction is submitted, but since it has Sync it is queued. Then if another task for this reaction comes in then that second task will be dropped (because one is already queued). Single influences how things run, Sync influences how things are queued.

src/book/02-system/01-nubots/03-nuclear.mdx
```diff
@@ -5,3 +5,144 @@ title: NUClear
...
+| Every    | Used to request periodic reactions in the system. `on<Every<2, std::chrono::seconds>()` will create a task every two seconds. `on<Every<2, Per<std::chrono::seconds>>()` will create a task two times per second.                                                                                                                                                                                                                              |
+| Always   | Used to request continuous reactions in the system. `on<Always>` will ensure a single instance of the associated reaction is running at all times.                                                                                                                                                                                                                                                                                             |
```
TrentHouliston: Maybe mention that on Always runs in its own thread rather than using the thread pool

src/book/02-system/01-nubots/03-nuclear.mdx
```diff
@@ -5,3 +5,144 @@ title: NUClear
...
+| Network  | NUClear provides a networking protocol to send messages to other devices on the network. `on<Network<T>>()` will create a task when T is emitted using the emission type `Scope::NETWORK`.                                                                                                                                                                                                                                                                 |
+
+More information on DSL words for on statements can be found in the [NUClear documentation](https://nuclear.readthedocs.io/en/latest/dsl.html#on-statements).
+
+### Runtime arguments
+
+Some DSL words will provide the ability to make changes to the system during runtime. This means that NUClear avoids the need for a system restart should a configuration, port number, or file need to be changed while the system is running.
+
+The DSL words that take runtime arguments are: `IO`, `TCP`, and `UDP`.
```
TrentHouliston: Since this is a NUbots specific page, we should probably also include `Configuration` and also include it in the dsl words above

src/book/02-system/01-nubots/03-nuclear.mdx
```diff
@@ -5,3 +5,144 @@ title: NUClear
...
+
+Emit statements are used by Reactors wanting to emit data to the PowerPlant. When using NUClear, data will most likely be emitted during a reaction. However, where necessary, emissions can also occur during reactor construction, or in some cases from within the PowerPlant itself.
+
+Any data emitted to the PowerPlant will be sent with a unique pointer. The PowerPlant will take ownership of this pointer and run any necessary callbacks to trigger reactions.
+
+Data can be emitted under varying scopes. These can be local or network emissions. More can be found on emit statements in the [NUClear documentation](https://nuclear.readthedocs.io/en/latest/dsl.html#emit-statements).
+
+### Local Emissions
+
+Local emissions send data to the local instance of the NUClear PowerPlant. Essentially this is the current running process. There are a number of scopes under which these emissions can take place. The syntax for these emissions are `emit<Scope::SCOPE>(data, dataType);`, where `SCOPE` is replaced with a scope as defined below.
```
TrentHouliston: There is no dataType what you might be referring to is extra arguments...
```suggestion
Local emissions send data to the local instance of the NUClear PowerPlant. Essentially this is the current running process. There are a number of scopes under which these emissions can take place. The syntax for these emissions are `emit<Scope::SCOPE>(data, args...);`, where `SCOPE` is replaced with a scope as defined below.
```

src/book/02-system/01-nubots/03-nuclear.mdx
```diff
@@ -5,3 +5,144 @@ title: NUClear
...
+
+| Scope      | Description                                                                                                                                  |
+| ---------- | -------------------------------------------------------------------------------------------------------------------------------------------- |
+| LOCAL      | When emitting data under this scope, tasks are distributed via the thread pool for execution. This is the default behaviour when emitting.   |
+| DIRECT     | When emitting data under this scope, the tasks created as a result of this emission will bypass the threadpool, and be executed immediately. |
+| INITIALISE | This scope emits data as the system starts up.                                                                                               |
+| DELAY      | This will emit data, after the provided delay. The syntax includes a delay time `emit<Scope::DELAY>(data, delay(ticks), dataType);`.         |
```
TrentHouliston: There is no dataType on the end
```suggestion
| DELAY      | This will emit data, after the provided delay. The syntax includes a delay time `emit<Scope::DELAY>(data, delay(ticks));`.         |
```

src/book/02-system/01-nubots/03-nuclear.mdx
```diff
@@ -5,3 +5,144 @@ title: NUClear
...
+
+### Network Emitting
+
+Network emissions can be used to send data through the network on which the current system is running.
+
+| Scope   | Description                                                                                                                          |
+| ------- | ------------------------------------------------------------------------------------------------------------------------------------ |
+| UDP     | Emits data as a UDP packet over the network. Is of the form `emit<Scope::UDP>(data, to_addr, to_port);`.                             |
+| NETWORK | Emits data over the network to other NUClear environments. Is of the form `emit<Scope::NETWORK>(data, target, reliable, dataType);`. |
```
TrentHouliston:
```suggestion
| NETWORK | Emits data over the network to other NUClear environments. Is of the form `emit<Scope::NETWORK>(data, target, reliable);`. |
```

src/book/02-system/01-nubots/03-nuclear.mdx
```diff
@@ -5,3 +5,144 @@ title: NUClear
...
+Let's say that type A is emitted to the PowerPlant, but at this time, type B does not have any data associated with it.
+
+Even though type B is missing data, since it's listed as optional, the task associated with this reaction can be scheduled. When executing the callback for the reaction, NUClear will identify that type B is not present, and will remove reference to this type from the callback, so that the task is only run with type A.
+
+Effectively, through the application of fission, the callback is restructured to look like the following example.
+
+```cpp
+.then([](const A& a){
+
+});
```
TrentHouliston: This isn't really how Fission works and also you'll still get a `b` it's just that if you try to use it you'll get a segmentation fault. I'll go put a more thorough explanation about how optional works and how you should use it up near that DSL word.

What fission actually is, is a way to split up function calls into multiple inputs and is mainly involved in emit from the point of view of the user. So for example, if you call `emit<Scope::NETWORK, Scope::DELAY>(message, "nusight", std::chrono::seconds(1));` the "fission" part, is that NUClear will send the `"nusight"` argument to the network emit and `std::chrono::seconds(1)` to the delay part. (not this won't combine them and network emit in 1 second, it'll just do an emit over the network **AND** a local emit in 1 second.

Closer to what you're describing here is an application of fusion and `RelevantArguments` that is used to ignore output from complicated DSL statements.
For example, when you do `on<Network<Data>, With<Sensors>>`, the network DSL word will make two arguments `const NetworkSource&, const Data&` and the with will make 1 `const Sensors&` but you can call it with `on<Network<Data>>().then([this](const Data& d, const Sensors& s) {});` which will "drop" the NetworkSource argument. This can be useful when you actually don't care about the contents of a message, or don't care about the source of a network packet etc. The matching is a greedy matching with each argument taking as much as it can before moving on.

src/book/02-system/01-nubots/03-nuclear.mdx
```diff
@@ -5,3 +5,144 @@ title: NUClear
...
+| disable()        | Disables the reaction. Tasks will not be created if triggered.                                                                                                                                      |
+| enable()         | Enables the reaction so that associated tasks will be scheduled and queued when the reaction is triggered.                                                                                          |
+| enable(bool set) | Sets the run status of the reaction handle. `true` will enable the reaction and `false` will disable it.                                                                                                   |
+| enabled()        | Determines the current status of the reaction. Returns `true` if the reaction is enabled, and `false` otherwise.                                                                                                           |
+| unbind()         | Removes a reaction request from the runtime environment. This action is not reversible, once a reaction has been unbound, it is no longer available for further use during that runtime. |
+
+# Examples
```
TrentHouliston: We should probably have a section on Neutrons as well and how they work with messaging

src/book/02-system/01-nubots/03-nuclear.mdx
TrentHouliston: Here is the explanation of how Optional works/should be used.

When you have an argument in NUClear it's not actually a const reference of that type, it is actually an std::shared_ptr<const T> object. NUClear just has the ability to work out if it could dereference an object to satisfy the type of the argument. So an on statement can look like either of these options.

```cpp
on<Trigger<DataType>>().then([this] (std::shared_ptr<const DataType> msg) {});
// OR
on<Trigger<DataType>>().then([this] (const DataType& msg) {});
```
The second one is easier to use most of the time, as when you're not using optional, NUClear won't run when you have no data.
However the times you do want to use the first one is when you're using Optional, because at those times that second shared_ptr argument could be a nullptr, and if you don't have a pointer you can't check if it's null or not. However if you have another way to check if that secondary object is null you can use the second option.

An example of the usage in the NUbots codebase is here
https://github.com/NUbots/NUbots/blob/master/module/platform/darwin/SensorFilter/src/SensorFilter.cpp#L248
and checking if it's null here
https://github.com/NUbots/NUbots/blob/master/module/platform/darwin/SensorFilter/src/SensorFilter.cpp#L314

The other time that the first option with the shared_ptr is used is when we want to save the pointer somewhere for future use as NUClear will only maintain it so long as it's the most recent object of that type. Although in that case since we use neutrons we often just use shared_from_this() instead, e.g.
https://github.com/NUbots/NUbots/blob/master/module/vision/GreenHorizonDetector/src/GreenHorizonDetector.cpp#L161
But in cases like that, we could have also just taken in a shared_ptr rather than a const reference

src/book/02-system/01-nubots/03-nuclear.mdx
```diff
@@ -5,3 +5,144 @@ title: NUClear
...
+
+| Scope      | Description                                                                                                                                  |
+| ---------- | -------------------------------------------------------------------------------------------------------------------------------------------- |
+| LOCAL      | When emitting data under this scope, tasks are distributed via the thread pool for execution. This is the default behaviour when emitting.   |
+| DIRECT     | When emitting data under this scope, the tasks created as a result of this emission will bypass the threadpool, and be executed immediately. |
+| INITIALISE | This scope emits data as the system starts up.                                                                                               |
+| DELAY      | This will emit data, after the provided delay. The syntax includes a delay time `emit<Scope::DELAY>(data, delay(ticks));`.         |
+
+### Network Emitting
```
JosephusPaye: To match **Local Emissions** above.
```suggestion
### Network Emissions
```

src/book/02-system/01-nubots/03-nuclear.mdx
ysims: I've changed the callback section if you can have a look. I'm not sure what should be added there, since that section seemed to just talk a lot about the optional word, which is addressed later.

src/book/02-system/01-nubots/03-nuclear.mdx
ysims: What are Neutrons?

src/book/02-system/01-nubots/03-nuclear.mdx
ysims: I've updated it a lil bit. I'll add this example to the examples section. Let me know if there are any other particularly good examples of anything NUClear-y in the codebase that you can think of.

ysims:
@JosephusPaye in the examples section I've got [`DarwinSensors`](https://github.com/NUbots/NUbots/blob/master/shared/message/platform/darwin/DarwinSensors.proto) as code and link but it only shows in code format. I notice GitHub makes it obvious that it is both code and link, can we get something similar for NUbook?

JosephusPaye:
That would be a good addition. Something like this:

image

ysims:
> That would be a good addition. Something like this:
> 
> 

Yes! That looks good

PR #46, opened 2020-03-05
Remove stray image files from reorganisation

JosephusPaye:
This removes left-over image files from the reorganisation in #32, and also renames the Guides section from number 4 to number 3.


JosephusPaye:
The tranzorb images are quite large (3-5MB) and should probably be resized. But maybe we can do that in a separate PR?

ysims:
I plan to do a sort of 'maintenance' PR soon on existing content (update info like meeting time, improve language, etc), I can look at all the images then and fix up size issues?

JosephusPaye:
> I plan to do a sort of 'maintenance' PR soon on existing content (update info like meeting time, improve language, etc), I can look at all the images then and fix up size issues?

Yep, that sounds good!

PR #57, opened 2020-04-25
Fix Prettier format error

JosephusPaye:
This fixes the following errors:

image

Currently Prettier's `mdx` parser [is broken](https://github.com/prettier/prettier/issues/7041), and fails when trying to parse JSX code blocks.

This is why code like this result in an error:

````mdx
<Alert>

```js
alert('oh hai')
```

</Alert>
````

Strangely, there's no error when using the `markdown` parser. So this PR adds a parser override to force Prettier to use the `markdown` parser for `.mdx` files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,15 @@
     "quoteProps": "consistent",
     "trailingComma": "es5",
     "jsxBracketSameLine": false,
-    "endOfLine": "lf"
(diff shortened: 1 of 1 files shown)


JosephusPaye:
Is this the same error you had @ElderNoSpace?

PR #69, opened 2020-09-30
Add Bibtex referencing

JosephusPaye:
By @moosahassanx, @samtarino94, and myself.

This PR adds a Bibtex referencing system to NUbook, much like the one in Overleaf.

### Tasks

- [x] Add Gatsby transformer to parse bib files and make their content available to pages in React
- [x] Add the `<cite>` element to show inline citations, with an error when a citation used is not found in the bib file
- [x] Add the `<References>` element
- [x] Use the IEEE referencing style (or whatever our Faculty uses at the uni)
  - [x] Format different reference types differently
- [x] Update NUbook page on how to use with examples

### Extensions (for future PRs)

- [ ] Add a popup to show the reference inline when a citation is hovered (like Wikipedia)
- [ ] Add a back link in the references list to jump back to where the citation was clicked
- [ ] Detect URLs in references and turn them into clickable links

### Preview

https://deploy-preview-69--nubook.netlify.app/kitchen-sink/referencing

### Docs

https://deploy-preview-69--nubook.netlify.app/system/tools/nubook#bibtex-referencing


src/components/markdown/references/references.jsx
```diff
@@ -0,0 +1,214 @@
...
+  // looping array
+  const referenceOptions = Array.from(usedReferences).map((referenceId, i) => {
+
+    const fields = references[referenceId].fields     // fields handles all the data you need, e.g. author, title, year, etc
+    const types = references[referenceId].type        // types handles the type of bibtex reference entry
+
+    const authorName = fields.author.split(' ');    // splitting author name for future use
+
+    // APA STYLING
+    if(citationStyle == "APA")
```
JosephusPaye, line 21: We can probably do with just one citation style: IEEE seems like a good option. [This article](https://guides.lib.monash.edu/citing-referencing/ieee) shows how to do it. [And this](https://vimeo.com/220916942) is also a nice video.

So we can just have the one style and remove the others. Keep the separate format for each type of reference, like you've done below with APA, it's nice 👍

src/components/markdown/references/references.jsx
```diff
@@ -0,0 +1,214 @@
+import React, { useContext } from 'react'
+import PropTypes, { array } from 'prop-types'
+
+import { BibReferencesContext } from '../../page-template'
+
+const References = ({ children, citationStyle }) => {
```
JosephusPaye, line 6: The `children` prop can be removed since it's not used, and `citationStyle` too (see my comment about a single format below).

```suggestion
const References = () => {
```

src/components/markdown/references/references.jsx
```diff
@@ -0,0 +1,214 @@
+import React, { useContext } from 'react'
+import PropTypes, { array } from 'prop-types'
```
JosephusPaye, line 2: Could remove the `array` import, since it's not used.

```suggestion
import PropTypes from 'prop-types'
```

src/components/markdown/references/references.jsx
```diff
@@ -0,0 +1,214 @@
+import React, { useContext } from 'react'
```
JosephusPaye, line 1: It seems like prettier is complaining about the format on this file.

You can run `yarn format` to fix it and then make a commit of the changes.

src/components/markdown/references/references.jsx
```diff
@@ -0,0 +1,214 @@
...
+    }
+  })
+
+  // return the map
+  return <div>{referenceOptions}</div>
+}
+
+References.PropTypes = {
+  children: PropTypes.isRequired,
+}
```
JosephusPaye, line 212: These won't be needed once the `children` and `citationStyle` props are removed.

```suggestion
```

JosephusPaye (commented):
This is a good start 👍.

Just needs a few changes and perhaps some custom styling, which I can help with.

JosephusPaye:
This should be ready to review and merge now.

JosephusPaye:
This should be ready to review and merge now.

ysims (commented):
Some comments on the styling, but these are just opinions.
1. From a user point of view, I'm not keen on actually having unknown references shown on the page. Is it possible to just throw an error? I don't see any benefit to showing it, and I was confused for a couple of seconds when I saw the first reference because I initially interpretted it as 'the first reference in the list is an unknown reference' before realising it was two separate references next to each other.
2. I don't know if I really like the way the inline citations look. After looking at it in dark mode, I think it's just that the darkness of the background in light mode feels a bit too noticable. Maybe it'd be better to have a light background with dark text like the references list at the bottom?

ysims (commented):
Styling looks good. Looked through some of the code and didn't see anything concerning but I can't say I know js well enough to comment. 
@lach-j can you check out the js? If that's good and no one has any other issues then I think it can be approved.

ysims:
@JosephusPaye probs also make an issue for those extensions.

lach-j (approved):
Looks good to me

PR #72, opened 2020-10-07
Prepare for search

JosephusPaye:
This PR adds some things we need in master before search can be added

- [x] Add sitemap
- [x] Update markup for Algolia


ysims (approved):
LGTM - excited for search!

PR #74, opened 2020-10-08
Formats NUsight page

ysims:
Somehow this slipped through, probably due to incorrect rules for the master branch which is now fixed (hopefully).


JosephusPaye:
LGTM. I think it slipped through because it was from a suggestion on GitHub (not pushed, which is when the formatting check happens).

JosephusPaye (approved):

JosephusPaye:
Updated the Netlify build.

PR #76, opened 2020-11-04
Fix anchor links

JosephusPaye:
This PR fixes anchor links (links that scroll to a specific section on the page).

It was broken due to the way we focused the active link in the sidebar.

Compare [this link](https://nubook.nubots.net/system/foundations/mathematics#velocity) (master, broken) with [the same link here](https://deploy-preview-76--nubook.netlify.app/system/foundations/mathematics#velocity) (this PR, fixed). Both links should open the Mathematics page and scroll to the Velocity section, but only the second works.

### DIFF
--- a/src/components/layout.jsx
+++ b/src/components/layout.jsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useRef } from 'react'
 import PropTypes from 'prop-types'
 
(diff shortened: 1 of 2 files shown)


ysims (approved):
Nice one! Looks good.

PR #78, opened 2020-11-20
Updates the Getting Started Guide

ysims:
Adds more information to the Getting Started guide.

- [x] Windows 10 Home
- [x] NUsight instructions for Ubuntu
- [x] NUsight instructions for Windows
- [x] NUsight instructions for MacOS
- [x] Follow an example all the way through (keyboardwalkfake)

[Preview.](https://deploy-preview-78--nubook.netlify.app/guides/main/getting-started)

### DIFF
--- a/src/book/03-guides/01-main-codebase/01-getting-started.mdx
+++ b/src/book/03-guides/01-main-codebase/01-getting-started.mdx
@@ -175,10 +175,11 @@ To build the code, open a terminal and change into the NUbots code directory. Th
 2. With the target platform selected, choose the binaries (called "roles") that you want to build. You do this by running the following command:
 
    ```bash
+   ./b configure
    ./b configure -i
    ```
 
-   This will show a menu to choose roles to build. Enable the roles you want to build and disable the ones you don't by highlighting them using using the <kbd>↑</kbd> and <kbd>↓</kbd> arrow keys and pressing the spacebar to toggle on and off.
+   Note that you will need to do `configure` without `-i` the very first time you do this. Using `-i` will show a menu that will allow you to choose the roles to build. Enable the roles you want to build and disable the ones you don't by highlighting them using using the <kbd>↑</kbd> and <kbd>↓</kbd> arrow keys and pressing the spacebar to toggle on and off.
 
    Next, press <kbd>c</kbd> to configure and then <kbd>g</kbd> to generate the build list and exit the menu.
 
--- a/src/book/03-guides/01-main-codebase/01-getting-started.mdx
+++ b/src/book/03-guides/01-main-codebase/01-getting-started.mdx
@@ -67,13 +67,19 @@ To install on Ubuntu, use the [official Docker instructions](https://docs.docker
 
 2. Complete the three steps under the heading 'Install Docker Engine', which is just below the previous section, 'Set Up the Repository'. After you complete this step, you are finished with the Docker install page.
 
-3. Add the current user to the Docker group, so you can run Docker commands without `sudo`.
+3. Install python3 and pip.
+
+   ```bash
+   sudo apt-get install python3 python3-pip
+   ```
+
+4. Add the current user to the Docker group, so you can run Docker commands without `sudo`.
 
    ```bash
    sudo usermod -aG docker "${USER}"
    ```
 
-4. Reboot to make the group change take effect.
+5. Reboot to make the group change take effect.
 
 </details>
 
--- a/src/book/03-guides/01-main-codebase/01-getting-started.mdx
+++ b/src/book/03-guides/01-main-codebase/01-getting-started.mdx
@@ -318,7 +318,7 @@ To run the code on a robot, you need to make sure the robot is powered on, insta
    ./robocup
    ```
 
-## NUsight
+## Get, Build and Run NUsight
 
 NUsight is the debugging application used at NUbots that visually represents what the robot is doing. It can be used to view output from any platform, whether that is a real NUgus robot or a simulation. Read more on the capabilities of NUsight on the [NUsight page](/system/tools/nusight).
 
@@ -395,61 +395,9 @@ Set the messages you would like NUsight to receive to 'true'. For our example, y
 
 <details>
 
-<summary> MacOS </summary>
-
-1. Get NUsight and cd into the directory by running
-
-   ```bash
-   git clone https://github.com/NUbots/NUsight2.git NUsight
-   cd NUsight
-   ```
-
-2. Create the Docker nuclearnet network to allow communication between the NUsight Docker container and the NUbots Docker container.
-
-   ```bash
-   docker network create nuclearnet
-   ```
-
-3. Make the Docker script executable by running
-
-   ```bash
-   chmod +x ./docker
-   ```
-
-4. Install the NUsight dependencies in the container and build NUsight for prod in the container
-
-   ```bash
-   ./docker yarn
-   ./docker yarn build
-   ```
-
-5. Inspect the network to find the NUbots container's IP address
-
-   ```bash
-   docker inspect --format='{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' nubots_generic
-   ```
-
-   The IP address given will be referred to as `<ADDRESS>`. You will need to be running a binary in NUbots or this will not work. See the info bubble below.
-
-   <Alert type='info'>
-
-   The IP address for the Docker container will not exist unless you are run a binary in it. See [Running with a Simulated Robot](#running-with-a-simulated-robot) for instructions.
-
-   </Alert>
-
-6. Run NUsight with the NUbots container address
-
-   ```bash
-   ./docker yarn prod --address <ADDRESS>
-   ```
-
-7. Open up in your browser the link that is given as output in the terminal, [localhost:9090](localhost:9090).
-
-</details>
-
 <details>
 
-<summary> Windows </summary>
+<summary> Windows/MacOS </summary>
 
 1. Get NUsight and cd into the directory by running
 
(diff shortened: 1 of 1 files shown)


src/book/03-guides/01-main-codebase/01-getting-started.mdx
```diff
@@ -26,15 +30,19 @@ There is a guide to using Git in NUbots [here](/guides/general/git).
 
 Docker is a container system that allows you to install development tools and run builds in an isolated environment that won't affect the rest of your system. We use it to set up the NUbots build system and run builds and binaries.
 
-#### Ubuntu
+You will need to perform the steps correctly to set up Docker for your operating system to complete the example.
+
+<details>
+
+<summary>Ubuntu</summary>
```
KipHamiltons, line 37: Since we specify `focal`, should we change this heading to `Ubuntu 20.04`?

Also, this section has headings for Mint, so should that be included in the heading too? We are assuming that these people won't know to even click on this tab, because they won't know the relationship between mint and ubuntu.

src/book/03-guides/01-main-codebase/01-getting-started.mdx
```diff
@@ -26,15 +30,19 @@ There is a guide to using Git in NUbots [here](/guides/general/git).
...
 
    ```bash
    curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
    sudo add-apt-repository \
      "deb [arch=amd64] https://download.docker.com/linux/ubuntu \
-     bionic \
+     focal \
      stable"
    sudo add-apt-repository universe
    sudo apt update
```
KipHamiltons, line 48: I think these two update commands should be added before the install commands (also, couldn't the install commands be merged into one command?), because I think mint users will run these commands, then it won't work, then they'll run the mint command, then they'll run the install commands, which won't work because these commands were ran before the special mint command.
I can see myself falling victim to this, because usually I am an idiot who doesn't read the next steps before running the current ones lmao

src/book/03-guides/01-main-codebase/01-getting-started.mdx
```diff
@@ -113,25 +125,56 @@ Docker is a container system that allows you to install development tools and ru
...
+
+</details>
 
 ### A Text Editor
 
+This section is not required for our example, but it is a good step to do in case you need to edit code in the future.
+
 You'll need a text editor to make changes to the code. We recommend [Visual Studio Code](https://code.visualstudio.com/Download), but feel free to use your favourite editor.
 
 To ensure a consistent [code style](https://en.wikipedia.org/wiki/Programming_style), NUbots uses `clang-format`, a tool that reformats all code files in the repository to have the same style, even if they're written by different people. For convenience, you can install the `clang-format` [extension](https://marketplace.visualstudio.com/items?itemName=xaver.clang-format) in Visual Studio Code to automatically format the files when you make changes.
```
KipHamiltons, line 180: While we're here... (lmk if you think this should be in a separate PR or if you disagree with it)
```suggestion
To ensure a consistent [code style](https://en.wikipedia.org/wiki/Programming_style), NUbots uses [clang-format](https://clang.llvm.org/docs/ClangFormat.html), a tool that reformats code files in the repository to have the same style, even if they're written by different people. For convenience, you can install the `clang-format` [extension](https://marketplace.visualstudio.com/items?itemName=xaver.clang-format) in Visual Studio Code to automatically format the files when you make changes.
```

src/book/03-guides/01-main-codebase/01-getting-started.mdx
```diff
@@ -175,24 +220,31 @@ To build the code, open a terminal and change into the NUbots code directory. Th
...
 
    ```bash
+   ./b configure
    ./b configure -i
    ```
 
-   This will show a menu to choose roles to build. Enable the roles you want to build and disable the ones you don't by highlighting them using using the <kbd>↑</kbd> and <kbd>↓</kbd> arrow keys and pressing the spacebar to toggle on and off.
+   Note that you will need to do `configure` without `-i` the very first time you do this. Using `-i` will show a menu that will allow you to choose the roles to build. Enable the roles you want to build and disable the ones you don't by highlighting them using using the <kbd>↑</kbd> and <kbd>↓</kbd> arrow keys and pressing the spacebar to toggle on and off.
+
+   If you are following the keyboardwalkfake example, you will need to make sure `keyboardwalkfake` is set to `true`. Find the list of items that are prefixed with `ROLE_` and set all to `false` except for `keyboardwalkfake`, which should be `true`. Do not change any other lines unless you know what you are doing.
```
KipHamiltons, line 229:
```suggestion
   If you are following the keyboardwalkfake example, you will need to make sure `keyboardwalkfake` is set to `true`. Find the list of items that are prefixed with `ROLE_` and set all to `false` except for `keyboardwalkfake`, which should be `true`. Do not change other lines unless you want to build a different role.
```

src/book/03-guides/01-main-codebase/01-getting-started.mdx
```diff
@@ -235,10 +303,197 @@ To run the code on a robot, you need to make sure the robot is powered on, insta
 
    This will start the binary to run the stand script, but the robot won't move until you press the green middle button on its back.
```
KipHamiltons, line 304: Should we mention what the red button does too?

KipHamiltons (commented):
Things past my last comment are things which I have no experience with, since I have never used nusight at all, so I didn't comment on them.

src/book/03-guides/01-main-codebase/01-getting-started.mdx
ysims: This was more about the build options than the other roles

src/book/03-guides/01-main-codebase/01-getting-started.mdx
ysims: Probably worth mentioning for safety reasons.

src/book/03-guides/01-main-codebase/01-getting-started.mdx
ysims: I've added info on what to change this too if you aren't using focal fossa. Also changed the drop-down heading to Ubuntu/Linux Mint.

src/book/03-guides/01-main-codebase/01-getting-started.mdx
ysims: Do you mean after? I've split up the commands into different steps, is it better?
On the one command thing, is that with `sudo add-apt-repository universe && sudo apt update` do you mean or something else? I'd rather not concatenate commands incase errors come up between commands.

KipHamiltons: I think where you have put step 2:
```sh
sudo add-apt-repository universe
sudo apt update
``` 
is perfect now.

The commands I was thinking could be merged would be 
```sh
sudo apt-get install docker-ce docker-ce-cli containerd.io
sudo apt-get install python3 python3-pip
```
becomes:
```sh
# Pretty sure this would work?
sudo apt-get install docker-ce docker-ce-cli containerd.io python3 python3-pip
```

Sidenote: that step says "Install the lastest version of Docker", then it proceeds to install docker, docker's dependencies (fair), and python too. Could perhaps just change it to saying something which encompasses both?

src/book/03-guides/01-main-codebase/01-getting-started.mdx
KipHamiltons: ahh true. I was just thinking that this might spook people a bit much, without telling them what they need to be spooked about. This means that they might know everything they need to know, but they wouldn't know that they did, so they might still be spooked (I am having a hard time articulating this in a way that makes sense...).

Kind of like how we frame `./b shell`, as something you [should basically never do](https://nubook.nubots.net/system/foundations/build-system#shell), even though it's quite handy a lot of the time. I have no idea why it's such a big deal, and since the documentation about it doesn't say why we should be careful, I still don't know if I am missing something. Although, I have used it a bunch, and it's really handy.

src/book/03-guides/01-main-codebase/01-getting-started.mdx
KipHamiltons: Looks like they have changed the steps to install Docker, so we'll need to add a few commands. I tried to test the combined command in a fresh ubuntu container, and now the steps we have listed won't work because we haven't added the GPG key for the repo...

https://docs.docker.com/engine/install/ubuntu/

Gonna have to add a couple of steps.

Something that's really nice though, is that they have added a command to the part where you change between focal and bionic so that we don't have to hardcode the version at all any more. Might not need divergent mint and ubuntu steps now.

KipHamiltons: On second thought, I reckon we could just change the guide to say "Go to Docker(link) and follow the steps to install it." at this stage (maybe with a comment saying that mint users should just follow the ubuntu steps).
That would make that `apt-get install <packages>` just `sudo apt-get install python3 python3-pip`

ysims: If it's `$(lsb_release -cs)`, I think that's always been a thing, it just doesn't work on Linux Mint.
Are the extra steps just 1 and 2 of 'SET UP THE REPOSITORY'?
We could add a link, kind of like an 'in case this doesn't work check if the instructions have changed' kind of thing. But there's a lot of extra stuff on there, and multiple ways to do things... so probs good to still have instructions here.

src/book/03-guides/01-main-codebase/01-getting-started.mdx
ysims: Haha yeah sure. Maybe just something like 'You do not need to change any other build options'.

src/book/03-guides/01-main-codebase/01-getting-started.mdx
KipHamiltons: The extra steps involve installing 5 more packages, then curling and adding the docker GPG key, checking the fingerprint to make sure you did those properly, then the steps we already have. I am not sure how long that fingerprint would be the same, so I am sceptical that we might have to update that regularly

ysims: Yeah maybe we should just link to it then. Then specify what section to do. And add any notes, like how to find what Ubuntu distro your Linux Mint version is.

KipHamiltons:
Really happy with how you've done this all. One minor change and it'll be good to go:
During the Docker consolidation, the python and pip installation fell of the radar. If we add them back in, I reckon it's good to go.

src/book/03-guides/01-main-codebase/01-getting-started.mdx
JosephusPaye: Good ol Kip the [rider](https://en.wikipedia.org/wiki/Rider_(legislation)) 😅.

ysims:
I think I'm happy with it now.

KipHamiltons (approved):
LGTM

PR #81, opened 2020-11-28
Add gazebo install instructions

thomas16441:
Add instructions for installing and running Gazebo using WSL

Note: Instructions are only for running using Ubuntu in windows, should probably include and format in a way that has instructions for Linux, Windows, MacOs separately

[Preview](https://deploy-preview-81--nubook.netlify.app/guides/tools/gazebo-install)


abihall (approved):
Great Instructions, they all worked and easy to follow

ysims (approved):
Looks fine for now, can do a separate PR for Linux and macOS in the future when we get around to it.

PR #99, opened 2021-02-26
Removes `wget` from arch flash guide

KipHamiltons:
wget was removed from newer arch iso's, so this replaces it with curl.

[preview](https://deploy-preview-99--nubook.netlify.app/guides/hardware/flashing#flashing-a-robot)


ElderNoSpace (approved):
LGTM

PR #100, opened 2021-03-18
Adds troubleshooting for dodgy cmake cache issue Daniel had today.

KipHamiltons:
[Preview](https://deploy-preview-100--nubook.netlify.app/guides/general/troubleshooting#why-am-i-having-cmakecache-errors-when-i-do-b-run-role)


Bidski (changes requested):
I would think the better option here would be to `./b configure --clean` and then `./b build`. Deleting `CMakeCache.txt` will effectively force a re-configure anyway and cleaning the docker build volume will more thoroughly resolve such errors (even if it maybe forces you to rebuild more than what just deleting `CMakeCache.txt` would).

`./b shell` is not a course of action that I think should be used on any regular basis, especially when other tools exist. To my mind `./b shell` is the fallback you use when all other options fail.

KipHamiltons:
We tried `./b configure --clean`, as well as changing the value for `USE_ASAN` in `./b configure -i`. Nothing worked except manually deleting that cache file

Bidski:
Are you sure the `CMakeCache.txt` wasnt in the project root? That has been known to cause some really bizarre problems (such as this one).

If it truly was in the build volume, and cleaning that volume didnt work then we need to work out why that is.

TrentHouliston:
I agree, it should be literally impossible to have this error provided you've built the code already. But if you do have a CMakeCache.txt in your project root directory it will cause these kinds of strange issues.

KipHamiltons:
Ahhh that was probably it. When I went in with `./b shell` I did a `find` which returned the location of two CMakeCache files, one of which was the normal one in the build folder, and the other one was in the repo somewhere (can't remember where exactly, but it wasn't the CM740 CMakeCache) and I just deleted them both without thinking about it.

I'll change this PR so that it tells you to look in your local repo and delete it instead of deleting the one in the container :+1:

src/book/03-guides/04-general/02-troubleshooting.mdx
```diff
@@ -35,6 +35,31 @@ Run `./b configure --clean`. This refreshes the docker volumes, removing the off
 
 If you have just added a new file to the codebase, make sure that it's part of a role. After you have added it to a role, run `./b configure`, and the build system should detect the new file and compile it the next time you run `./b build`.
 
+### Why am I having CMakeCache errors when I do `./b run <role>`?
```
Bidski, line 38: Having a `CMakeCache.txt` file in the NUbots repo can also cause bizarre issues with `./b install` as well. There may also be other issues that this can cause. It might be good to add a note letting people know that if they are getting weird errors with other b-script modules that looking for the `CMakeCache.txt` file may potentially resolve it

src/book/03-guides/04-general/02-troubleshooting.mdx
```diff
@@ -35,6 +35,31 @@ Run `./b configure --clean`. This refreshes the docker volumes, removing the off
...
+    use_asan = b.cmake_cache["USE_ASAN"] == "ON"
+KeyError: 'USE_ASAN'
+```
+
+If the problem persists, you probably have an extra `CMakeCache.txt` file in your NUbots directory, which is messing up the build's cache.
+Fixing this issue is as simple as finding the offending file and deleting it.
+You can find the file in your text editor, or by running the find command inside the NUbots directory:
+
+```sh
+find ./ -name "CMakeCache.txt"
```
Bidski, line 58:
```suggestion
find . -type f -name CMakeCache.txt -delete && find . -type d -name CMakeFiles -delete
```

src/book/03-guides/04-general/02-troubleshooting.mdx
```diff
@@ -35,6 +35,31 @@ Run `./b configure --clean`. This refreshes the docker volumes, removing the off
...
+  File "/home/nubots/NUbots/tools/utility/dockerise/run.py", line 79, in _run
+    func(**kwargs)
+  File "/home/nubots/NUbots/tools/run.py", line 34, in run
+    use_asan = b.cmake_cache["USE_ASAN"] == "ON"
+KeyError: 'USE_ASAN'
+```
+
+If the problem persists, you probably have an extra `CMakeCache.txt` file in your NUbots directory, which is messing up the build's cache.
+Fixing this issue is as simple as finding the offending file and deleting it.
+You can find the file in your text editor, or by running the find command inside the NUbots directory:
```
Bidski, line 55: I would do this first. The existence of this file (and the `CMakeFiles` folder) is indicative of a larger problem and removing these will likely resolve other issues that are being experienced. If removing these doesnt fix anything, then I would purge the build volume (`./b configure --clean`)

Bidski (changes requested):

KipHamiltons:
@Bidski I combined your suggestions and rehashed the changes :+1:

Bidski (approved):
That seems better

PR #108, opened 2021-05-28
Update webots guide to use Robocup TC fork, add instructions to change mirrors

JosephusPaye:
This changes the Webots guide to install the Robocup TC's fork of Webots, and adds instructions to change the webots download mirrors to use [our version on GitHub](https://github.com/NUbots/WebotsDependenciesMirror).

With this change installs are a lot faster (downloads take minutes instead of hours). 

[Preview](https://deploy-preview-108--nubook.netlify.app/guides/tools/webots-setup#install-webots)

### DIFF
--- a/src/book/03-guides/02-tools/03-webots.mdx
+++ b/src/book/03-guides/02-tools/03-webots.mdx
@@ -67,11 +67,11 @@ Follow the rest of the steps on this page within Ubuntu.
 
 </details>
 
-Install libraries for cmake, protobuf, eigen, yaml-cpp, and ninja-build:
+Install libraries for cmake, protobuf, eigen, yaml-cpp, ninja-build, and clang-tidy:
 
 ```sh
 sudo apt-get update
-sudo apt-get install cmake-curses-gui libprotobuf-dev protobuf-compiler libeigen3-dev libyaml-cpp-dev ninja-build
+sudo apt-get install cmake-curses-gui libprotobuf-dev protobuf-compiler libeigen3-dev libyaml-cpp-dev ninja-build clang-tidy
 ```
 
 ## Install Webots
--- a/src/book/03-guides/02-tools/03-webots.mdx
+++ b/src/book/03-guides/02-tools/03-webots.mdx
@@ -80,12 +80,20 @@ sudo apt-get install cmake-curses-gui libprotobuf-dev protobuf-compiler libeigen
 
    ```sh
    cd ~
-   git clone --branch develop --single-branch --recurse-submodules -j$(nproc) https://github.com/cyberbotics/webots.git
+   git clone --branch release --single-branch --recurse-submodules -j$(nproc) https://github.com/RoboCup-Humanoid-TC/webots.git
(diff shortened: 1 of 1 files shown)


KipHamiltons:
I think we could provide instructions for if you have already got `webots` installed. Something like this:
```bash
cd $WEBOTS_HOME
git remote set-url origin https://github.com/RoboCup-Humanoid-TC/webots.git
git fetch
git checkout release
# `make cleanse` removes the deps, which we want to keep!
# `make clean` just removes the compiled stuff, which we will need to do
make clean
make -j$(nproc)
```
should do it I think. I am pretty sure if you already had it installed, you don't need to redownload the deps/apply your patch at all.
If you think recloning is easier though, I respect that, especially since you mirrored the deps lol.


We will also need to add `clang-tidy` to the list of NUWebots deps to install by `apt-get install`, so might as well do that here.

src/book/03-guides/02-tools/03-webots.mdx
```diff
@@ -80,12 +80,20 @@ sudo apt-get install cmake-curses-gui libprotobuf-dev protobuf-compiler libeigen
...
-2. Follow the rest of the instructions on the [Webots GitHub repository](https://github.com/cyberbotics/webots/wiki/Linux-installation). Follow these instructions inside Ubuntu (WSL) if you are a Windows user. See the previous section on how to set this up. The optional dependencies are not required.
+2. Change the download mirrors for Webots depedencies by running the following:
+
+   ```sh
+   cd webots
+   wget https://github.com/NUbots/WebotsDependenciesMirror/raw/master/change-mirrors.patch
+   git apply change-mirrors.patch
+   ```
+
+3. Follow the rest of the instructions on the [Webots GitHub repository](https://github.com/cyberbotics/webots/wiki/Linux-installation). Follow these instructions inside Ubuntu (WSL) if you are a Windows user. See the previous section on how to set this up. The optional dependencies are not required.
```
ysims, line 96: This should say something about where to start from with Cyberbotic's instructions, since it shouldn't be the full page (not cloning anyway)

ysims: Although I think we should follow these instructions? https://github.com/RoboCup-Humanoid-TC/webots/blob/release/projects/samples/contests/robocup/README.md

ysims: https://github.com/NUbots/webots/ I want to add config for our team in this fork then we can easily use it without ppl not knowing what the config should be.

wongjoel (approved):
Ran through the steps on my computer, and it worked

src/book/03-guides/02-tools/03-webots.mdx
JosephusPaye: Updated instructions to use ours (copied into NUbook).

JosephusPaye:
Merging now, we can add the stuff about updating an existing install later. For now, remove + re-clone is the way to go.

PR #73, opened 2020-10-08
Adds a maintenance guide page with information on maintaining vision

ysims:
Maintenance for vision includes

- Dataset generation
- Training and testing the Visual Mesh (off-robot) and putting it into the NUbots main codebase
- Camera calibration (more just linking to the page, but still waiting for that to go into master)
- Testing the output of vision on the robot
- Tuning detectors

[Preview.](https://deploy-preview-73--nubook.netlify.app/guides/main/maintaining-subsystems)

PR #111, opened 2021-06-03
Add tutorial walking through training the visual mesh with webots data

wongjoel:
This is a practical guide designed to walk a newish member of the team through how to generate a training dataset in Webots, then train the visual mesh on that dataset, then run the newly visual mesh. The main aim is to preserve the knowledge of how to train and run the visual mesh so that it is common knowledge for vision team members. This may also be a good learning tool for people who prefer to learn by following a tutorial rather than reading an explanation of how something works.

This is mostly a placeholder to remind us to finish this after Robocup 2021

Preview: NULL

### DIFF
--- a/src/book/03-guides/02-tools/05-visualmesh-webots.mdx
+++ b/src/book/03-guides/02-tools/05-visualmesh-webots.mdx
@@ -1,0 +1,29 @@
+---
+section: Guides
+chapter: Tools
+title: Visual Mesh with Webots
+description: Getting Data from webots to train the Visual Mesh.
+slug: /guides/tools/visualmesh
+---
+
+This guide walks through the process of generating a training dataset using Webots and using that dataset to train the Visual Mesh.
+
+<Alert type='info'>
+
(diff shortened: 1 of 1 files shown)


ysims:
We need to be careful with the amount of pages we have. It is perfectly fine to have a long page with 5-7 headings. Having a sidebar with 50 pages is not fine.

I think this would fit into the page being created in #73, in the section on getting datasets.

LachlanCourt:
@wongjoel What was the decision around this? I think content wise the tutorial is fine although choosing where to put it was the issue?

wongjoel:
To be honest, I'd kind of forgotten about this: my current thought is that perhaps this project should be geared around revamping the [existing guide](https://nubook.nubots.net/guides/tools/visualmesh) that uses a dummy dataset. Perhaps the existing guide could be extended to mention how to generate a webots vision dataset (instead of dummy), and also extended to mention how to check if the mesh is actually working.

PR #122, opened 2021-06-27
Add a page describing the director algorithm

TrentHouliston:
[Preview](https://deploy-preview-122--nubook.netlify.app/system/foundations/director)


src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+```
+
+### When
+
+A when condition is used to ensure that a provider will only execute when its conditions are met.
+These conditions are provided using a global enum message along with a statement that describes what condition you want.
+For example if you wanted a kick engine to only run if the robot is statically stable or better.
+
+```cpp
+on<Provides<Kick>, When<Stability, std::greater_equal, Stability::STATIC>>().then(...);
```
ysims, line 116: When feels long. But I guess there's no better way. Something like `<Stability >= STATIC>` might be good but potentially restricting.

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+### Causing
+
+A causing statement is a promise that by running this provider, the desired state will be eventually reached.
+This is used to allow smooth transitions beween modules.
+For example when the walk engine stopped moving it would be in a "Standing" stability state.
+This means the walk engine could have a `Leaving` reaction that caused this state so other modules know it can use this to be standing before they execute their actions.
+
+```cpp
+on<Leaving<Walk>, Causing<Stability, Stability::STANDING>().then(...);
```
ysims, line 129: This Causing looks nicer than that When above. So I guess it might not be too long in the end.

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+```
+
+Other providers will take advantage of this in two ways.
+If the other provider has higher priority than this one, it will force this provider group to switch to a `Provides` or `Leaving` that has a causing that satisfies the when condition it is limited on.
+If the other provider has lower priority than this one, it will only be able to force this provider to change to another `Provides` statement and not a leavign one.
```
ysims, line 134:
```suggestion
If the other provider has lower priority than this one, it will only be able to force this provider to change to another `Provides` statement and not a leaving one.
```

ysims: Right so if a kick has a higher priority to the walk then it can force the walk to change to the leaving provider that causes standing/stability.

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+## Tasks
+
+Tasks are the jobs that should be executed using the providers.
```
ysims, line 138: So only providers emit tasks? Wait do all reactors become providers or is this wrong thinking...

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+Tasks are implemented using NUClears `emit` along with a custom scope `emit<Task>`.
+
+### Root level tasks
+
+When a task is emitted from a reaction which is not a provider, that task is considered a root level task.
```
ysims, line 143: Oh I guess this answers my question about. But I'm a little confused on the boundary with providers and other reactors.

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+These root level tasks are all considered siblings within the priority comparision system at the root level of the graph.
```
ysims, line 144: What does this mean? They all occur at the same time?

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+When multiple tasks are emitted by a provider, each of these are sibling tasks in the director graph.
+
+### Priority
+
+Priority in the Director is considered based on the closest common anscestor.
+For root tasks this will be the root element.
+Once the closest common anscestor is determined, the priority of the branch we are on is compared to the priority of the branch of the other task.
+The winner gets to use the resources and potentially evict the other task to wait in the queue for the provider group.
+
+Tasks can also be emitted as optional tasks, if when finding a common anscestor between two tasks, an optional task is found in the parantage of a task before the common anscestor that task is considered "optional".
```
ysims, line 159:
```suggestion
Tasks can also be emitted as optional tasks, if when finding a common ancestor between two tasks, an optional task is found in the parantage of a task before the common anscestor that task is considered "optional".
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+A subtask is a task that is emitted by a provider.
+When multiple tasks are emitted by a provider, each of these are sibling tasks in the director graph.
+
+### Priority
+
+Priority in the Director is considered based on the closest common anscestor.
+For root tasks this will be the root element.
+Once the closest common anscestor is determined, the priority of the branch we are on is compared to the priority of the branch of the other task.
+The winner gets to use the resources and potentially evict the other task to wait in the queue for the provider group.
```
ysims, line 157: I feel like I need a picture for this whole thing or more in depth explanation.

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+Tasks can also be emitted as optional tasks, if when finding a common anscestor between two tasks, an optional task is found in the parantage of a task before the common anscestor that task is considered "optional".
+If one task has an optional parentage and the other does not then the optional task will automatically lose.
+If both have optional tasks in their perantage, then the tasks will be compared using their common anscestor as normal as if the optional was not there.
+This means the higher priority root tasks optional elements are more important than a lower priority tasks optional elements.
+
+### Done tasks
+
+A Done task is a special task that can be emitted by a provider to signal to its parent that it has completed the task that was assigned to it.
```
ysims, line 166: Like the get up/scripts telling everyone it's finished?

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+The provider group that created this task will then be re-executed with the knowledge that it was triggered as a done event from below.
+If a root task emits a `Done` task then it will be removed from the Director's root tasks as if it were emitted with a priority of 0.
+
+## The Director Graph
+
+TODO algorithm development in progress
```
ysims, line 172: Exciting.

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+## Behaviours Layout
```
ysims, line 174: About this whole section - so all of these do become providers? Webots, HardwareIO, QuinticWalk, etc? This is a complete codebase wide change? This almost feels like NUClear 2.0 but maybe I think it does more than what it really does because I don't understand it properly.

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+Leaving providers are used for the state when a provider group is losing control and will no longer have access to subtasks, or when a higher priority task elsewhere needs a causing that it provides.
+Leaving providers are executed in two scenarios:
+
+Firstly, A leaving provider will be executed once when the provider no longer has an active task.
+Note: If there no leaving providers that are satisfied in this situation (due to when conditions) the system will stop executing the provider without informing it.
+It may be advantageous to have a `Leaving` provider with no when conditions if you want to always know when you no longer have a task.
+
+Secondly, a leaving provider may be executed as a result of it having a causing reaction that another task needs if it has a higher priority than this task.
+For example, if a higher priority kick engine wanted to kick being staticallys stable, and the walk engine had a leaving which caused statically stable, then the leaving .
```
ysims, line 97:
```suggestion
For example, if a higher priority kick engine wanted to kick being statically stable, and the walk engine had a leaving which caused statically stable, then the leaving .
```

ysims: You didn't finish your sentence.

ysims (commented):
This is interesting. Many comments.

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+## Providers
+
+Providers are implemented as NUClear reactions and live in a NUClear `on` statement.
+They are the components which "provide" the functionality needed to perform a task, either by doing it directly or by running a series of subtasks.
+They form the basis of the modules that implement behaviours.
+Each provider will provide for a specific message type in a specific situation.
+For example a provider which executed a walk task, the code might look like:
+
+```cpp
+on<Provides<Walk>>().then(...);
```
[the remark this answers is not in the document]
TrentHouliston: It's true, but I wasn't sure what else to put here and I wanted to put the exact code construct I was talking about somewhere

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+Provider are triggered whenever a task needs to be executed by the director, however they can also be triggered using other NUClear keywords.
+The director will ensure that any provider which is not currently active will not run if it is not active, even if the other triggers would make it run.
+For example if the walk engine module wants to update at 10Hz when it is running a provider may look like.
+
+```
+on<Provides<Walk>, Every<10, Per<std::chrono::seconds>>().then(...);
+```
+
+This provider would then run once every time the walk command was updated, as well as running at 10Hz while the provider is active.
```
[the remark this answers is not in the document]
TrentHouliston: The way it disables providers is not so much like calling .disable() on the reaction, it will be using the same mechanism where if you have an on trigger with a with, and the with hasn't been emitted yet it won't run.

So really when you have something else (like an every) running on a provider that the director is not using, it's more like there isn't a task there for it to access so it doesn't run.

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+If you do want a provider to run even when it is inactive you can make them optional (e.g. `on<Optional<Provides<Walk>>>().then(...);`).
```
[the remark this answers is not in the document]
TrentHouliston: I can't think of a use case for Optional, but I thought I would mention it in case someone else does.

There are a few ways of dealing with things like getup, my personal preference is just a root-level task that gets high priority when the robot has fallen over, which will override any strategy or anything else and get up

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+Then this will still run although its task will be empty.
+Non active providers are unable to emit subtasks and will show an error if they attempt to.
+
+Note that for the director to recognise subtasks as coming from a specific provider, those tasks MUST be emitted from that specific reaction.
+Otherwise they will be considered root tasks and can cause all sorts of chaos.
+
+### Provider Groups
+
+When there are multiple providers which provide for a single task type they are considered a provider group.
```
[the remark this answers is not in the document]
TrentHouliston: So multiple providers for a single task has heaps of use cases, those you suggested can surely be one if you want it to be.
The more general case I was thinking of is when you have things like a normal walk engine provider
`on<Provides<Walk>>`
And another provider that makes the robot stand by stopping walking
`on<Leaving<Walk>, Causing<Stability::STAND>`

Both of those providers take a walk command and get the walk engine to do something.
When you want to run either of those depends on what you're doing.
If a higher priority task wants to run then it can take advantage of these other providers to get the robot into a state that it wants to (e.g. stand before kick) without you having to program in specialised state control variables to ensure the robot has stopped walking before the kick happens

You can take this concept super far though by making multiple providers that run in different circumstances depending on how crazy you wan to go

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+When a new task is emitted the director will determine which of the providers in a group to run based on the current state of the system as well as any extra DSL keywords that are on the providers decleration.
+The director may also update which provider in a group is running if a state update means that the currently active provider can no longer run.
```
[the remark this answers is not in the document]
TrentHouliston: You could do something like this, if you have a way of monitoring some state.

E.g. you could have a walk where it's
`on<Provides<Walk>, When<SomeCondition, std::less, SomeCondition::State>>`
and another where it's
`on<Provides<Walk>, When<SomeCondition, std::less, SomeCondition::State2>>`

and if both of them succeed with the current state, but the first one doesn't succeed with the second ones state, it'll prefer the second one. So you can have teirs of things if you really want.

I'd advise that you save more advanced usages of the director like that for after you get a basic method happening.
But it should work in theory without you having to put much programming into it

When I finish the director algorithm I intend for it to always choose the most specific provider that matches given whens (the one that the fewest other providers match).

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+In this case it will attempt to reassign that task to a provider which is able to run it.
```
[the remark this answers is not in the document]
TrentHouliston: If there is no provider at all, that's a bug in the code and the director will yell at you.
For example if you try to run walk, but don't have a walk module installed it will yell at you.

On the other hand, if there is just no module that can run because of a state, like kick wants to run but the state doesn't match, Then the director will try to do things that will allow the kick to run. Or if it can't find one, it'll just wait and hope and run whatever the next highest priority thing is.

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+Provider groups operate together as a single unit, so any task that is emitted from a provider in a group is considered as beign emitted from that group.
+So if you are running one Provider, and the director changes you to another provider, if that provider emits different subtasks than the original the subtasks that were not emitted will be considered "finished" and removed from the graph.
```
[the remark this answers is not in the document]
TrentHouliston: This one is especially useful for the higher level strategy stuff, where you want to do something and then change to wanting to do something else.

Like if there is some global provider that monitors the game state, and changes which behaviours it's doing like emit PenalisedBehaviour when penalised, so all the other actions in the system get dumped and it just does a penalised behaviour. Then emit a PenaltyShootoutBehaviour later and it'll cascade and do all that

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+An entering provider is used to set up the state of a provider when it is running, or to ensure that a When condition of a main provider is met.
+They are executed in two scenarios:
+
+1. when a provider group has no active tasks and then is given to one.
+2. when it was running a task normally but the only `Provides` reaction that is running has its when condition fail and needs to re-enter to provide a `Causing`.
+
+Example:
+
+```cpp
+on<Entering<Walk>>().then()
```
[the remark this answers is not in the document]
TrentHouliston: I did want to make a full example but I wasn't sure how deep to go :P

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+Example:
+
+```cpp
+on<Provides<Walk>>().then()
+```
+
+### When
+
+A when condition is used to ensure that a provider will only execute when its conditions are met.
+These conditions are provided using a global enum message along with a statement that describes what condition you want.
```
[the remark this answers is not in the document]
TrentHouliston: Syncs everywhere, so hopefully no race conditions in the Director.
Although depending on the exact circumstances it is possible to bounce a state on an off, and confuse the director as it will enable/disable tasks.

But to be honest the director is doing the right thing and the state is just unstable

src/book/02-system/01-foundations/07-director.mdx
TrentHouliston: If I could find a way to make it so you could do Stability >= STATIC I would do it.
if you can find a way to make that code work so I can get that as an expression and not a value I can make it behave like that.

In the meantime you get the advantage that the std::greater_equal can actually be any function if you want to get crazy with it (probably not a good idea)

src/book/02-system/01-foundations/07-director.mdx
TrentHouliston: Correct, that way the kick doesn't have to care about "how" it's getting to the state it needs. It just needs to know it's the most important thing to do right now and the director will try to find a way to make it work.

So that means no hacky things in the kick engine or another module where you monitor when the walk has finished walking so you can activate the kick etc, it'll just work itself out

src/book/02-system/01-foundations/07-director.mdx
TrentHouliston: Providers are reactions, so on statements become providers.
On statements with the same type in them form a provider group

src/book/02-system/01-foundations/07-director.mdx
TrentHouliston: Providers are just reactions with special properties as to how they run.
So a reactor will contain one or more providers

src/book/02-system/01-foundations/07-director.mdx
TrentHouliston: This means that when you're comparing the priority of the tasks, you compare against your first common ancestors children (ignoring optional).

So if you had two things emitted and they had a conflict later the one that would win would be the one with higher priority at the point where they diverged in the graph (the provider that emitted both of them or in this case all the other root tasks)

src/book/02-system/01-foundations/07-director.mdx
TrentHouliston: More like a servo command telling the script engine it finished it's current motion so the script engine knows to send the next command

src/book/02-system/01-foundations/07-director.mdx
TrentHouliston: I feel like there is a simpler way of explaining this. The code for it wasn't too hard.
Basically, find your common ancestor, work out which child you are and what the priority is compared to the other child you're competing with. These priorities only come up if two things want the same resources. E.g. walk and kick both want the legs then only one can have it. The one who came from the highest priority parent gets the legs.

Then optional just means if either is optional the other wins and if both are optional then just do the algorithm as normal

src/book/02-system/01-foundations/07-director.mdx
TrentHouliston: It is a fairly codebase wide change, but a lot of it will just be replacing things that were previously on<Trigger<X>> with on<Provides<X>> so that the director can choose when to run them.

Some of the modules will get smaller because of this and some will likely need some retooling and splitting up.

At some point when we want to start working on the modules I intend to make a list of all the modules I see currently in the codebase so people can start building and testing them.

Best part about the director is you can start with any task as a root level task which will make it super easy to test things.
No more of this weird subsumption registration so you can get PS3 walk going. PS3 walk would just emit a walk command as a root level task and update it.

src/book/02-system/01-foundations/07-director.mdx
TrentHouliston: I should finish my sentence

src/book/02-system/01-foundations/07-director.mdx
TrentHouliston: Something something provider groups making a graph via edges made by tasks, and tasks that can't run due to low priority or when conditions get thrown into a queue in the provider group.

Is current algorithm progress

src/book/02-system/01-foundations/07-director.mdx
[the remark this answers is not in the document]
ysims: Oh right. So you could disable these reactors as well as deactivating the providers?

src/book/02-system/01-foundations/07-director.mdx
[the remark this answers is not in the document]
ysims: Right but still "Do all in a group have to have some sort of priority in case of an unknown order?"
How does it decide which one if it isn't obvious? and they're siblings?

src/book/02-system/01-foundations/07-director.mdx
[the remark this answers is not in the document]
ysims: The NUClear page I think has decent examples. It's good I think. I can only process so much in English.

src/book/02-system/01-foundations/07-director.mdx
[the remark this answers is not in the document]
ysims: Sounds good to me. Bouncing is just something for the users of it to look out for I think.

src/book/02-system/01-foundations/07-director.mdx
ysims: Sounds good.

src/book/02-system/01-foundations/07-director.mdx
ysims: Oh so that is a yes that all reactors will have providers?

src/book/02-system/01-foundations/07-director.mdx
ysims: Oh right that exact scenario would fix the same thing hacky hacksaws fixed (badly) in the last few hours of comp.

src/book/02-system/01-foundations/07-director.mdx
ysims: Sounds good. I'm wondering how we can make the change incrementally without having a huge PR that causes major merge conflicts for every existing branch. Will this also change how we write roles?

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
+---
+section: System
+chapter: Foundations
+title: Director
+description: Introduction to the Director and how to use it to build behaviours.
+slug: /system/foundations/director
+---
+
+The NUbots use the director system to control all their behaviour elements.
```
LachlanCourt, line 9:
```suggestion
The NUbots use the Director system to control all their behaviour elements.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+It then uses these keywords to make a graph that it will use to decide which tasks to execute in order to achive the goals of the modules.
+
+The two DSL components that are added are Providers and Tasks.
+
+## Providers
+
+Providers are implemented as NUClear reactions and live in a NUClear `on` statement.
+They are the components which "provide" the functionality needed to perform a task, either by doing it directly or by running a series of subtasks.
+They form the basis of the modules that implement behaviours.
+Each provider will provide for a specific message type in a specific situation.
```
LachlanCourt, line 22:
```suggestion
Each Provider will provide for a specific message type in a specific situation.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+For example a provider which executed a walk task, the code might look like:
```
LachlanCourt, line 23:
```suggestion
For example a Provider which executed a walk task, the code might look like:
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+```cpp
+on<Provides<Walk>>().then(...);
+```
+
+Provider are triggered whenever a task needs to be executed by the director, however they can also be triggered using other NUClear keywords.
```
LachlanCourt, line 29:
```suggestion
Providers are triggered whenever a task needs to be executed by the Director, however they can also be triggered using other NUClear keywords.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+The director will ensure that any provider which is not currently active will not run if it is not active, even if the other triggers would make it run.
```
LachlanCourt, line 30:
```suggestion
The Director will ensure that any Provider which is not currently active will not run if it is not active, even if the other triggers would make it run.
```

I am confused here - are you saying it will make sure it won't try to run the same Provider twice, simultaneously?

src/book/02-system/01-foundations/07-director.mdx
LachlanCourt, line 37:
```suggestion
This Provider would then run once every time the walk command was updated, as well as running at 10Hz while the Provider is active.
```

src/book/02-system/01-foundations/07-director.mdx
LachlanCourt, line 39:
```suggestion
If you do want a Provider to run even when it is inactive you can make them `Optional` (e.g. `on<Optional<Provides<Walk>>>().then(...);`).
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+```
+on<Provides<Walk>, Every<10, Per<std::chrono::seconds>>().then(...);
+```
+
+This provider would then run once every time the walk command was updated, as well as running at 10Hz while the provider is active.
+
+If you do want a provider to run even when it is inactive you can make them optional (e.g. `on<Optional<Provides<Walk>>>().then(...);`).
+Then this will still run although its task will be empty.
+Non active providers are unable to emit subtasks and will show an error if they attempt to.
```
LachlanCourt, line 41:
```suggestion
Non active Providers are unable to emit subtasks and will show an error if they attempt to.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+Note that for the director to recognise subtasks as coming from a specific provider, those tasks MUST be emitted from that specific reaction.
```
LachlanCourt, line 43:
```suggestion
Note that for the Director to recognise subtasks as coming from a specific Provider, those tasks MUST be emitted from that specific reaction.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+Otherwise they will be considered root tasks and can cause all sorts of chaos.
```
LachlanCourt, line 44: All sorts of chaos is my favourite kind of chaos but this sentence might need to be reworked :D

src/book/02-system/01-foundations/07-director.mdx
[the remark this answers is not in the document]
LachlanCourt:
```suggestion
When there are multiple Providers which provide for a single task type they are considered a provider group.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+### Provider Groups
+
+When there are multiple providers which provide for a single task type they are considered a provider group.
+When a new task is emitted the director will determine which of the providers in a group to run based on the current state of the system as well as any extra DSL keywords that are on the providers decleration.
```
LachlanCourt, line 49:
```suggestion
When a new task is emitted the Director will determine which of the Providers in a group to run based on the current state of the system as well as any extra DSL keywords that are on the Provider's declaration.
```

src/book/02-system/01-foundations/07-director.mdx
LachlanCourt, line 50:
```suggestion
The Director may also update which Provider in a group is running if a state update means that the currently active Provider can no longer run.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+The director may also update which provider in a group is running if a state update means that the currently active provider can no longer run.
+In this case it will attempt to reassign that task to a provider which is able to run it.
+
+Provider groups operate together as a single unit, so any task that is emitted from a provider in a group is considered as beign emitted from that group.
```
LachlanCourt, line 53:
```suggestion
Provider groups operate together as a single unit, so any task that is emitted from a Provider in a group is considered as being emitted from that group.
```

src/book/02-system/01-foundations/07-director.mdx
LachlanCourt, line 54:
```suggestion
So if you are running one Provider, and the Director changes you to another Provider, if that Provider emits different subtasks than the original, the subtasks that were not emitted will be considered "finished" and removed from the graph.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+So if you are running one Provider, and the director changes you to another provider, if that provider emits different subtasks than the original the subtasks that were not emitted will be considered "finished" and removed from the graph.
+
+### Provider Types
+
+There are three different provider types that can be used.
```
LachlanCourt, line 58:
```suggestion
There are three different Provider types that can be used.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+These three provider along with their `When` and `Causing` DSL words determine the flow of a provider, executing code as the director changes which tasks it executes.
```
LachlanCourt, line 59:
```suggestion
These three Providers along with their `When` and `Causing` DSL words determine the flow of a Provider, executing code as the Director changes which tasks it executes.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+#### Entering
+
+An entering provider is used to set up the state of a provider when it is running, or to ensure that a When condition of a main provider is met.
```
LachlanCourt, line 63:
```suggestion
An `Entering` Provider is used to set up the state of a Provider when it is running, or to ensure that a `When` condition of a Main Provider is met.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+They are executed in two scenarios:
+
+1. when a provider group has no active tasks and then is given to one.
```
LachlanCourt, line 66:
```suggestion
1. When a Provider group has no active tasks and then is given to one.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+2. when it was running a task normally but the only `Provides` reaction that is running has its when condition fail and needs to re-enter to provide a `Causing`.
```
LachlanCourt, line 67:
```suggestion
2. When it was running a task normally but the only `Provides` reaction that is running has its when condition fail and needs to re-enter to provide a `Causing`.
```

src/book/02-system/01-foundations/07-director.mdx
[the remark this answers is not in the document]
LachlanCourt: Give in to the c++ brain!!

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+Example:
+
+```cpp
+on<Entering<Walk>>().then()
+```
+
+#### Provides
+
+Provides are the normal providers which will be expected to execute a task.
+They will be chosen as the active provider in a group if there are no Entering statements in a provider group, or if at least one entering provider has executed making this group "entered".
```
LachlanCourt, line 78:
```suggestion
They will be chosen as the active Provider in a group if there are no `Entering` statements in a Provider group, or if at least one `Entering` Provider has executed making this group "entered".
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+However an entering provider may be repeatedly run still if all the providers have `When` conditions that are not yet met.
```
LachlanCourt, line 79:
```suggestion
However an `Entering` Provider may be repeatedly run still if all the Providers have `When` conditions that are not yet met.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+Example:
+
+```cpp
+on<Provides<Walk>>().then()
+```
+
+#### Leaving
+
+Leaving providers are used for the state when a provider group is losing control and will no longer have access to subtasks, or when a higher priority task elsewhere needs a causing that it provides.
```
LachlanCourt, line 89:
```suggestion
Leaving Providers are used for the state when a Provider group is losing control and will no longer have access to subtasks, or when a higher priority task elsewhere needs a `Causing` that it provides.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+Leaving providers are executed in two scenarios:
```
LachlanCourt, line 90:
```suggestion
`Leaving` Providers are executed in two scenarios:
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+Firstly, A leaving provider will be executed once when the provider no longer has an active task.
```
LachlanCourt, line 92:
```suggestion
Firstly, A `Leaving` Provider will be executed once when the Provider no longer has an active task.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+Note: If there no leaving providers that are satisfied in this situation (due to when conditions) the system will stop executing the provider without informing it.
```
LachlanCourt, line 93:
```suggestion
Note: If there no `Leaving` Providers that are satisfied in this situation (due to when conditions) the system will stop executing the Provider without informing it.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+It may be advantageous to have a `Leaving` provider with no when conditions if you want to always know when you no longer have a task.
+
+Secondly, a leaving provider may be executed as a result of it having a causing reaction that another task needs if it has a higher priority than this task.
```
LachlanCourt, line 96:
```suggestion
Secondly, a `Leaving` Provider may be executed as a result of it having a causing reaction that another task needs if it has a higher priority than this task.
```

src/book/02-system/01-foundations/07-director.mdx
LachlanCourt: I thought it was an artistic choice

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+For example, if a higher priority kick engine wanted to kick being staticallys stable, and the walk engine had a leaving which caused statically stable, then the leaving .
+In the event that the kick providers task had a lower priority than this providers task, then it is only able to change this module to run other non `Leaving` providers to achive its causing.
```
LachlanCourt, line 98:
```suggestion
In the event that the kick Provider's task had a lower priority than this Provider's task, then it is only able to change this module to run other non `Leaving` Providers to achive its causing.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+In order for a leaving provider to be run however, it must have first been in the entering or normal states.
```
LachlanCourt, line 100:
```suggestion
In order for a `Leaving` Provider to be run however, it must have first been in the entering or normal states.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+The system cannot immediately jump to a leaving provider.
```
LachlanCourt, line 101:
```suggestion
The system cannot immediately jump to a `Leaving` Provider.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+Example:
+
+```cpp
+on<Provides<Walk>>().then()
+```
+
+### When
+
+A when condition is used to ensure that a provider will only execute when its conditions are met.
```
LachlanCourt, line 111:
```suggestion
A `When` condition is used to ensure that a provider will only execute when its conditions are met.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+```cpp
+on<Provides<Kick>, When<Stability, std::greater_equal, Stability::STATIC>>().then(...);
+```
+
+These when conditions are updated by using normal NUClear emits.
+
+### Causing
+
+A causing statement is a promise that by running this provider, the desired state will be eventually reached.
```
LachlanCourt, line 123:
```suggestion
A `Causing` statement is a promise that by running this Provider, the desired state will be eventually reached.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+This is used to allow smooth transitions beween modules.
+For example when the walk engine stopped moving it would be in a "Standing" stability state.
+This means the walk engine could have a `Leaving` reaction that caused this state so other modules know it can use this to be standing before they execute their actions.
+
+```cpp
+on<Leaving<Walk>, Causing<Stability, Stability::STANDING>().then(...);
+```
+
+Other providers will take advantage of this in two ways.
```
LachlanCourt, line 132:
```suggestion
Other Providers will take advantage of this in two ways.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+If the other provider has higher priority than this one, it will force this provider group to switch to a `Provides` or `Leaving` that has a causing that satisfies the when condition it is limited on.
```
LachlanCourt, line 133:
```suggestion
If the other Provider has higher priority than this one, it will force this Provider group to switch to a `Provides` or `Leaving` that has a `Causing` that satisfies the `When` condition it is limited on.
```

src/book/02-system/01-foundations/07-director.mdx
LachlanCourt, line 134:
```suggestion
If the other Provider has lower priority than this one, it will only be able to force this Provider to change to another `Provides` statement and not a `Leaving` one.
```

src/book/02-system/01-foundations/07-director.mdx
LachlanCourt, line 138:
```suggestion
Tasks are the jobs that should be executed using the Providers.
```

src/book/02-system/01-foundations/07-director.mdx
LachlanCourt, line 143:
```suggestion
When a Task is emitted from a Reaction which is not a Provider, that task is considered a root level task.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+### Root level tasks
+
+When a task is emitted from a reaction which is not a provider, that task is considered a root level task.
+These root level tasks are all considered siblings within the priority comparision system at the root level of the graph.
+When a root level task is emitted with 0 priority it is removed from the list of tasks.
+
+### Subtasks
+
+A subtask is a task that is emitted by a provider.
```
LachlanCourt, line 149:
```suggestion
A Subtask is a Task that is emitted by a Provider.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+When multiple tasks are emitted by a provider, each of these are sibling tasks in the director graph.
```
LachlanCourt, line 150:
```suggestion
When multiple Tasks are emitted by a Provider, each of these are sibling tasks in the Director graph.
```

src/book/02-system/01-foundations/07-director.mdx
LachlanCourt, line 157:
```suggestion
The winner gets to use the resources and potentially evict the other task to wait in the queue for the Provider group.
```

LachlanCourt: How do we specify that one comes before the other?

src/book/02-system/01-foundations/07-director.mdx
LachlanCourt, line 159:
```suggestion
Tasks can also be emitted as optional Tasks, if when finding a common ancestor between two tasks, an optional Task is found in the parantage of a task before the common anscestor that Task is considered "optional".
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+### Priority
+
+Priority in the Director is considered based on the closest common anscestor.
+For root tasks this will be the root element.
+Once the closest common anscestor is determined, the priority of the branch we are on is compared to the priority of the branch of the other task.
+The winner gets to use the resources and potentially evict the other task to wait in the queue for the provider group.
+
+Tasks can also be emitted as optional tasks, if when finding a common anscestor between two tasks, an optional task is found in the parantage of a task before the common anscestor that task is considered "optional".
+If one task has an optional parentage and the other does not then the optional task will automatically lose.
```
LachlanCourt, line 160:
```suggestion
If one Task has an optional parentage and the other does not then the optional Task will automatically lose.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+If both have optional tasks in their perantage, then the tasks will be compared using their common anscestor as normal as if the optional was not there.
```
LachlanCourt, line 161:
```suggestion
If both have optional Tasks in their perantage, then the Tasks will be compared using their common anscestor as normal as if the optional was not there.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+This means the higher priority root tasks optional elements are more important than a lower priority tasks optional elements.
+
+### Done tasks
+
+A Done task is a special task that can be emitted by a provider to signal to its parent that it has completed the task that was assigned to it.
+The provider group that created this task will then be re-executed with the knowledge that it was triggered as a done event from below.
```
LachlanCourt, line 167:
```suggestion
The Provider group that created this Task will then be re-executed with the knowledge that it was triggered as a done event from below.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+If a root task emits a `Done` task then it will be removed from the Director's root tasks as if it were emitted with a priority of 0.
```
LachlanCourt, line 168:
```suggestion
If a root Task emits a `Done` task then it will be removed from the Director's root tasks as if it were emitted with a priority of 0.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+To the Director they are all equivalent but to make it easier for developers they are split into several different layers.
+
+### Behaviour
+
+Behavour modules are the highest level modules and during normal debugging these will be the primary tasks that will be executed by the system.
+They live in `modules/behaviour/behaviour` in the source code.
+The generally will describe the overall goals and behavour of the robot.
+Some examples of a behavour level module would be:
+
+- Striker: Combines stratigies to walk to the ball and kick it to the opposing goal
```
LachlanCourt, line 186:
```suggestion
- Striker: Combines strategies to walk to the ball and kick it towards the opposing goal
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+- Defender: Combines strategies needed to intercept and defend incoming goals
+- Goalie: Combines strategies needed to stay near the goal and prevents the opponents to score
+
+### Strategy
+
+Behaviours are modules that make high level strategy decisions for the robot.
```
LachlanCourt, line 192:
```suggestion
Strategies are modules that make high level strategy decisions for the robot.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+They live in `modules/behaviour/strategy` in the source code.
+These modules are the behaviour elements that provide the "what" components rather than the "how" components.
```
LachlanCourt, line 194:
```suggestion
These modules are the strategic elements that provide the "what" components rather than the "how" components.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+- WalkToBall: Tells the walk path planner to walk to the ball location
+- LookForBall: Tells the look planner to look for a specific set of objects
+
+### Planners
+
+Planners are where the system starts to think about the "how" part of behavour.
+They live in `modules/behaviour/planner` in the source code.
+Planner modules are modules which string together a set of skills in order to achive their goal.
+They are only interested in a single component of perofrming an action, such as walking a path or
```
LachlanCourt, line 206:
```suggestion
They are only interested in a single component of performing an action, such as walking a path or:
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+### Skills
+
+- Walk: The walk engine which walks given a walk command
+- Kick: Kicks the ball according to the desired direction
+- Look: Gets the head to look at a paticular location
+
+### Motions
+
+The motions layer is concenred with implmementing the primitive actions
```
LachlanCourt, line 219:
```suggestion
The Motions layer is concerned with implmementing the primitive actions
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+- ScriptRunner: Executes scripts given to it
+- InverseKinematics: Takes limb positions/head positions and converts them into servo positions
+
+### Platform
+
+The platform layer deals with sending commands to the hardware itself.
```
LachlanCourt, line 226:
```suggestion
The Platform layer deals with sending commands to the hardware itself.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+These modules should not have subtasks and will form the leaf nodes in the directors graph.
```
LachlanCourt, line 227:
```suggestion
These modules should not have subtasks and will form the leaf nodes in the Directors graph.
```
Leaf nodes?? First I'm hearing of this wizardry

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,231 @@
...
+title: Director
+description: Introduction to the Director and how to use it to build behaviours.
+slug: /system/foundations/director
+---
+
+The NUbots use the director system to control all their behaviour elements.
+It allows composable behaviours that can take advantage of transition rules to ensure that the robot performs the right task at the right time.
+
+The Director does this by providing a set of new DSL keywords to NUClear that describe what tasks to run, what conditions are required, and how those tasks build on other tasks.
+It then uses these keywords to make a graph that it will use to decide which tasks to execute in order to achive the goals of the modules.
```
LachlanCourt, line 13: More on this graph might be handy

yarn.lock
```diff
@@ -2,6 +2,19 @@
 # yarn lockfile v1
```
LachlanCourt, line 2: Is this supposed be here? I'm not 100% on how NUbook works but none of the other PRs have this file changed

src/book/02-system/01-foundations/07-director.mdx
TrentHouliston: I'd recommend doing it in modules from the bottom up.
So write the servo module
Then the IK module
Then the walk module
etc

ysims: Will this also change how we write roles?

yarn.lock
KipHamiltons: The `yarn.lock` file sets versions for your dependencies. It should be changed when you change your dependencies or their versions. More info [here](https://classic.yarnpkg.com/en/docs/yarn-lock/). 
Basically, good catch! It shouldn't be committed here.

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -324,8 +324,11 @@ Some examples are
...
-- ScriptRunner: Executes scripts given to it
-- InverseKinematics: Takes limb positions/head positions and converts them into servo positions
+These modules live in `modules/actuation` in the source code.
+The end of the actuation layer involves emitting servo commands to be picked up by a platform module, such as HardwareIO or Webots modules.
+Modules in this layer include
+
+- Kinematics: Takes end effector positions and converts them into servo positions
+- Sequences: Takes a sequence of servo positions and emits them in order
+- Groups: Takes a group of servos and emits them together, will only emit Done when all servos are done
+- Servos: Takes a single servo and emits it, will only emit Done when the servo has finished moving
```
TrentHouliston, line 334: Only when finished moving was an extension you wanted to add. Currently, it'll emit done when out of time.

ysims: This are just examples. I'm not sure if even having this list here is very good since it should be on the relevant subsystem page, and two places to have to update makes it more likely it'll get out of date.

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,447 @@
...
+
+Subtasks must be emitted from the Provider reaction to be considered a child of that Provider. If the Task is emitted from a non-Provider reaction, it will be considered a root task.
+
+### Provider Groups
+
+When there are multiple Providers which provide for a single Task type they are considered a provider group. For example, a both a static walk module and a ZMP walk module could have `Provide<Walk>` reactions. These two Providers would be considered a Provider group for the Walk task.
+
+When a Provider group is running, only one of the Providers in that group is allowed to execute at a time.
+
+When a new Task is emitted the Director will determine which of the Providers in a group to run based on the current state of the system as well as any extra DSL keywords that are on the Provider's decleration. If both Providers could run, then the one declared first will run.
```
LachlanCourt, line 58:
```suggestion
When a new Task is emitted the Director will determine which of the Providers in a group to run based on the current state of the system as well as any extra DSL keywords that are on the Provider's declaration. If both Providers could run, then the one declared first will run.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,447 @@
...
+on<Provide<Kick>, When<Stability, std::greater_equal, Stability::STANDING>>().then(...);
+```
+
+These conditions are updated by using normal NUClear emits. For example if
+
+```cpp
+emit(std::make_unique<Stability>(Stability::STATIC));
+```
+
+was emitted in another reaction, then the `When` condition would be met and the Provider would run.
```
LachlanCourt, line 141: This becomes a code block so either the second line should be capitalised or it should be an inline code snippet

```suggestion
Was emitted in another reaction, then the `When` condition would be met and the Provider would run.
```

LachlanCourt, line 141: This becomes a code block so either the second line should be capitalised or it should be an inline code snippet. I think the latter flows better but up to you

```suggestion
Was emitted in another reaction, then the `When` condition would be met and the Provider would run.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,447 @@
...
+Provider groups operate together as a single unit. Any Task that is emitted from a Provider in a group is considered as being emitted from that group. Considering the walk examples, both Providers may emit a `LegIK` Task. Regardless of which Provider is active, the `LegIK` Task will be considered as coming from the walk Provider group. If the walk Providers emit different subtasks, then the old subtasks will be considered "finished" and removed from the graph, and the new subtasks will start running.
+
+### Provider Types
+
+There are three different Provider types that can be used.
+These three provider types along with the `Needs`, `When` and `Causing` DSL words determine the flow of a Provider, executing code as the Director changes which Tasks it executes.
+
+#### Start
+
+A Start Provider is used to set up the state of a Provider when it starts running. Start is executed when a Provider group has no active Tasks and then is given to one, ie it gains control.
```
LachlanCourt, line 71:
```suggestion
A Start Provider is used to set up the state of a Provider when it starts running. Start is executed when a Provider group has no active Tasks and then is given to one, i.e. it gains control.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,447 @@
...
+    if (state == STOPPED) {
+        // Emit that we are standing
+        emit(std::make_unique<Stability>(Stability::STANDING));
+    }
+});
+```
+
+The implementation for the reaction should result in the desired state being eventually emitted. This would not be the primary Provider for the Walk, but would instead be actie when a higher priority Provider wants this state.
+
+For example, if the kick engine only runs `When<Stability, std::greater_equal, Stability::STANDING>` then it would not be able to run while the walk engine is running normally. However, if the kick Task has a higher priority than the walk Task, then it will push the walk engine into the Causing Provide reaction, which will then cause the Stability state to be STANDING so that the kick engine can run.
```
LachlanCourt, line 164:
```suggestion
For example, if the kick engine only runs `When<Stability, std::greater_equal, Stability::STANDING>` then it would not be able to run while the walk engine is running normally. However, if the kick Task has a higher priority than the walk Task, then it will push the walk engine into the Causing Provide reaction, which will then cause the Stability state to be `STANDING` so that the kick engine can run.
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,447 @@
...
+
+This walk engine Task contains information on the speed and direction that the robot should walk in. Other Tasks, such as `GetUpWhenFallen` does not need any data.
+
+Tasks can take four arguments
+
+```cpp
+emit<Task>(data, priority, optional, name)
+```
+
+where
```
LachlanCourt, line 185:
```suggestion
Where
```

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,447 @@
...
+
+- data: A pointer to a Protobuf message.
```
LachlanCourt, line 187: I think this might look nicer as a table maybe... Or at least put each of the params in backticks?

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,447 @@
...
+on<Provides<Walk>>().then([this] (const Walk& walk, const RunReason& reason) {
+    if (reason == RunReason::NEW_TASK) {
+        log<NUClear::INFO>("A new Walk Task triggered this reaction");
+    }
+});
+```
+
+Supported RunReasons are
+
+- NEW_TASK: The Provider is running because a new Task of the type it provides for was emitted.
```
LachlanCourt, line 285: Again maybe a table or backticks for this

LachlanCourt (changes requested):

src/book/02-system/01-foundations/07-director.mdx
ysims: I'll opt for making it inline because it's not a complete sentence otherwise

ysims, line 141:
```suggestion
These conditions are updated by using normal NUClear emits. For example if `emit(std::make_unique<Stability>(Stability::STATIC));` was emitted in another reaction, then the `When` condition would be met and the Provider would run.
```

src/book/02-system/01-foundations/07-director.mdx
ysims, line 185:
```suggestion
Tasks can take four arguments `emit<Task>(data, priority, optional, name)` where
```

bencoady (changes requested):

src/book/02-system/01-foundations/07-director.mdx
```diff
@@ -0,0 +1,442 @@
...
+| optional | A boolean. If true, then the Task is considered optional. If two Tasks need the same resources and have the same priority but one is optional, then the non-optional Task will be chosen. Optional Tasks do not need to be active for other Tasks from the same Provider to run. If no optional argument is specified, then the default optional value of false is used. |
+| name     | A string. If provided, then this string will be used when logging messages. The default is the empty string.                                                                                                                                                                                                                                                             |
+
+### Root Level Tasks
+
+When a Task is emitted from a reaction which is not itself a Provider, that Task is considered a root level task.
+These root level tasks are all considered siblings within the priority comparison system at the root level of the graph.
+When a root level task is emitted with a `nullptr` is removed from the list of Tasks.
```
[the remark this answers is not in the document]
ysims:
```suggestion
When a root level task is emitted with a `nullptr`, that root level task is removed from the Director tree and the list of Tasks to execute.
```
Not sure if this is better. The idea is that you can remove tasks in providers (ie non root level tasks) by just not emitting them when you next run the provider. But for root-level, there is no provider to do this from, since they originate outside any provider. The way to remove them from the director system is to emit the task again with a nullptr.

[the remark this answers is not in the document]
bencoady: Yeah that's definitely clearer.

bencoady (approved):

PR #131, opened 2021-07-04
Fixes heading levels so that they progress correctly

ysims:
Some pages started with level 1 headings (`#`), however this is technically incorrect. A level 1 heading is automatically generated on the page for the title of the page. Contributers to NUbook should start with level 2 headings (`##`) in markdown files. 

The NUpbr page had heading level 3 skipped, this was fixed so that the headers progressed correctly.


JosephusPaye (approved):
LGTM (I've also updated the Kitchen Sink pages, except for the uses that are intentional).

In the future we might be able to fix this automatically with `yarn format`.

PR #132, opened 2021-07-04
Replace `sh` with `bash` in terminal code blocks for proper syntax highlighting

JosephusPaye:
This will correctly highlight comments, command names, and other shell syntax.


ysims (approved):
Looks good, can't see any that are missed.

PR #138, opened 2021-07-13
Unit test guide and catch example

### LINKED ISSUE #213: Add section about pairing via bluetooth
Don't mind me. I'm just documenting the steps for connecting bluetooth controllers here. I'll add these in later with a pull request...
----
The instructions are based on the ones [located here](https://wiki.archlinux.org/title/Gamepad#Pairing_via_Bluetooth), except with one minor difference: depending on the robot there is a **large amount of interference**. This creates dead zones where the PS3 controller can't send data. The dead zones are very clear cut. Slight movements can enable the PS3 to send data.

One spot which works quite well is placing the controller up against the back of the robot's case, just below the power buttons.

A few other points:
1. Run `jstest --normal /dev/input/js0` to verify that the controller is sending data. `jstest`can be installed using `pacman -S joyutils` if it's not on the robot already. [See here](https://wiki.archlinux.org/title/Gamepad#Joystick_API).
3. It might be nice to enable auto power on boot for the bluetooth controller on the robot. [See here](https://wiki.archlinux.org/title/bluetooth#Auto_power-on_after_boot/resume).
4. The bluetooth service is not started by default. Use `systemctl start bluetooth` to start it, and `systemctl enable bluetooth` to start it when the robot powers on.

KipHamiltons:
Very cool and epic guides. One covers unit testing, another gives a little detail of a catch example.

TODO:

- [ ] Add stuff about catch BDD tests to catch guide, and try to bring focus onto that style

[Preview](https://deploy-preview-138--nubook.netlify.app/guides/tools/catch-guide)


src/book/03-guides/04-general/04-writing-tests.mdx
```diff
@@ -0,0 +1,60 @@
...
+---
+
+This guide presents some general information about unit testing and tests in general.
+
+## What is a Unit Test?
+
+A unit test is a test of a small piece of a codebase - a unit. Unit tests are meant to test only that piece of code and they serve to validate its correctness. This contrasts with integration tests, which tests the interactions of pieces of code and their behaviour together. For the rest of this guide, we'll work with the following toy utility function
+
+```cpp
+[[nodiscard]] inline int add(const int a, const int b) noexcept {
```
ysims, line 16: 😆 I can't help but wonder if this would confuse some people.

src/book/03-guides/04-general/04-writing-tests.mdx
```diff
@@ -0,0 +1,60 @@
...
+title: How to Write Tests
+description: Information About Making Tests
+slug: /guides/general/writing-tests
+---
+
+This guide presents some general information about unit testing and tests in general.
+
+## What is a Unit Test?
+
+A unit test is a test of a small piece of a codebase - a unit. Unit tests are meant to test only that piece of code and they serve to validate its correctness. This contrasts with integration tests, which tests the interactions of pieces of code and their behaviour together. For the rest of this guide, we'll work with the following toy utility function
```
ysims, line 13:
```suggestion
A unit test is a test of a small piece of a codebase - a unit. Unit tests should test a single piece of code to validate its correctness. This contrasts with integration tests, which tests the interactions of pieces of code and their behaviour together. For the rest of this guide, we'll work with the following toy utility function
```
Suggested change because I stumbled over the words a bit when reading it.

src/book/03-guides/04-general/04-writing-tests.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+### Regression Tests
+
+When we find bugs in the codebase and fix them, we should add a test case which makes sure that that code doesn't _regress_ into the buggy behaviour. This test case is called a regression test. Regression tests should be labelled with comments to indicate the behaviour they're watching for. If there is a GitHub issue related to the bug, the comment with the test should reference it. The test should be written such that it would fail before the fix, and pass after the fix.
+
+For a concrete example, imagine that a bug was found with `add` where if both inputs were negative, it always returned 0. Good practice would be to add a test case or small set of test cases where both inputs were negative - such as `(-1, -1) == -2` and `(-123, -456) == -579` - labelling them as regression tests for that bug. These would clearly fail if the bug came back (although this example is quite contrived, because there should have already been tests with both inputs negative).
```
ysims, line 50: I had to google 'contrived'. Either my vocabulary is lacking there or maybe better to use a different word, 'manufactured' maybe?

src/book/03-guides/04-general/04-writing-tests.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+### Black Box vs White Box Testing
+
+When we write tests, we can make them completely ignorant to the internals of the code. Such a test worries only about the inputs and outputs, considering the parts in the middle as a black box which we can't see inside - black box testing.
+
+White box testing looks at the internals and makes tests which depend on them. This means that any significant change to the implementation of a function which doesn't change its interface is likely to break white box tests which depended on the old function. The fragility of white box tests to change is the chief reason that black box tests are preferred.
+
+Ideally, you're able to write tests as you design your software interfaces, writing the code afterwards such that it fulfills the needs of the tests. This is the basis of [**Test-Driven Development**](https://en.wikipedia.org/wiki/Test-driven_development)(TDD), which is a powerful means of creating high quality software. Tests written as part of a TDD process should inherently be black box tests, because the implementations they're testing don't exist yet.
```
ysims, line 58:
```suggestion
Ideally, you're able to write tests as you design your software interfaces, writing the code afterwards such that it fulfills the needs of the tests. This is the basis of [**Test-Driven Development**](https://en.wikipedia.org/wiki/Test-driven_development) (TDD), which is a powerful means of creating high quality software. Tests written as part of a TDD process should inherently be black box tests, because the implementations they're testing don't exist yet.
```

src/book/03-guides/02-tools/05-catch-guide.mdx
```diff
@@ -0,0 +1,65 @@
...
+
+Catch test cases have a few components. The most important is the `TEST_CASE(..)` macro, which wraps around groups of associated tests. Inside each `TEST_CASE` scope, you should follow the **AAA** structure, _Arranging_ the data first - both inputs and expected outputs - then _Acting_ by calling the function you're testing, then _Asserting_ that the results match the ground truth. We'll make an example for the following toy utility function:
+
+```cpp
+[[nodiscard]] inline int add(const int a, const int b) noexcept {
+    return a + b;
+}
+```
+
+To test it, we'll need pairs of inputs and their associated outputs. Usually, if there's a lot of data, we'll want to keep the it separate from the test logic, but for this example we'll keep it local. Also note that for utilities like this one, we would want a much more comprehensive set of test cases.
```
ysims, line 23:
```suggestion
To test it, we'll need pairs of inputs and their associated outputs. Usually, if there's a lot of data, we'll want to keep it separate from the test logic, but for this example we'll keep it local. Also note that for utilities like this one, we would want a much more comprehensive set of test cases.
```

src/book/03-guides/02-tools/05-catch-guide.mdx
```diff
@@ -0,0 +1,65 @@
...
+        INFO("Ground truth is " << ground_truth[i]);
+        INFO("Function output is " << outputs[i]);
+        REQUIRE(outputs[i] == ground_truth[i]);
+    }
+}
+```
+
+### Dissecting the Example
+
+As we can see in this example, inside the scope of the `TEST_CASE` is where it all happens. We use `INFO` macros to commentate exactly what is happening for each assertion. You shouldn't worry too much about creating huge, unwieldy logs with `INFO` macros, because Catch only prints the `INFO` for test cases which fail by default.
```
ysims, line 51:
```suggestion
As we can see in this example, inside the scope of the `TEST_CASE` is where it all happens. We use `INFO` macros to commentate exactly what is happening for each assertion. You shouldn't worry too much about creating huge, unwieldy logs with `INFO` macros, because by default Catch only prints the `INFO` for test cases which fail.
```

ysims (commented):
It might be good to link to the catch guide in the how to write tests guide? I'd even maybe go as far as to say, combine the pages (see troubleshooting page)? Otherwise I'd also suggest that 'Catch Getting Started' should be under 'Main Codebase'. 
If I wanted to write tests and knew nothing, 'Catch Getting Started' wouldn't stand out to me as a page that would hold the answers I seek.

src/book/03-guides/04-general/04-writing-tests.mdx
JosephusPaye: Kip and his love of nodiscard inline noexcept 😅

src/book/03-guides/04-general/04-writing-tests.mdx
JosephusPaye: I like contrived, it fits here nicely

src/book/03-guides/04-general/04-writing-tests.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+### Regression Tests
+
+When we find bugs in the codebase and fix them, we should add a test case which makes sure that that code doesn't _regress_ into the buggy behaviour. This test case is called a regression test. Regression tests should be labelled with comments to indicate the behaviour they're watching for. If there is a GitHub issue related to the bug, the comment with the test should reference it. The test should be written such that it would fail before the fix, and pass after the fix.
+
+For a concrete example, imagine that a bug was found with `add` where if both inputs were negative, it always returned 0. Good practice would be to add a test case or small set of test cases where both inputs were negative - such as `(-1, -1) == -2` and `(-123, -456) == -579` - labelling them as regression tests for that bug. These would clearly fail if the bug came back (although this example is quite contrived, because there should have already been tests with both inputs negative).
+
+### Black Box vs White Box Testing
```
CMurtagh-LGTM, line 52:
```suggestion
## Black Box vs White Box Testing
```
Not a part of the anatomy :thinking:

CMurtagh-LGTM (dismissed):

src/book/03-guides/04-general/04-writing-tests.mdx
```diff
@@ -0,0 +1,60 @@
...
+    return a + b;
+}
+```
+
+## Anatomy of a Unit Test
+
+The basic pieces you will need to create a unit test for a function are the following:
+
+1. A set of inputs for the function. These should match the input parameters. In our case, we'll need a set of pairs of integers.
+2. The associated outputs which you want for each input. These are often referred to as the "ground truth", because they are the true values your function should output. For our `add` example with the pair of inputs (2, 2), we want the output 4, because `add(2, 2) == 4`.
```
Crafty15, line 26: Could add something about how we (prefer to) get the ground truth data. Using python scripts with comparable libraries in the case of testing math functions for example.

src/book/03-guides/04-general/04-writing-tests.mdx
```diff
@@ -0,0 +1,60 @@
...
+The process of running the tests is as simple as calling the function with your inputs and verifying that the function's output was as expected. In our example, one of the test cases could be running `add(2, 2)` and checking that the result is 4.
+
+## Testing Approach and Philosophy
+
+There are many guidelines you can find online to help you to write good tests. Here are a few:
+
+- Tests should be simple and readable enough to be correct on inspection. You don't want to think about whether a test is correct or not. Ideally you'll be able to read it and know that it's legitimate.
+- Make test cases independent. The outcome of one test case shouldn't affect the outcome of others.
+- Demonstrate how a piece of code should be used with its tests. We can't google for examples of people using our software, so create examples with your tests.
+- Tests should be deterministic - [seed your randomness](https://en.wikipedia.org/wiki/Random_seed). If you're testing something particularly reliant on randomness or which generates randomness, compensate by using a variety of seeds with many cases each.
```
Crafty15, line 37: What would be an example of something reliant on randomness?

src/book/03-guides/04-general/04-writing-tests.mdx
```diff
@@ -0,0 +1,60 @@
...
+- Follow the **AAA** structure: Arrange, Act, Assert. Each test should follow the general design of first setting up your input variables (Arrange), calling your unit with those variables (Act), then finally checking your outputs match what they should (Assert).
+
+### General Approach
+
+Write the easy tests first, then think about edge cases and code coverage. For `add`, you might make the `(2, 2) == 4` case first, then `(123000, 456) == 123456`. After you have some simple cases, you could consider throwing in some zeros and negative numbers - cases where the observed behaviour is different somehow.
+
+Later, you might consider what should happen on integer overflow/wraparound, making sure that errors are handled correctly. At that stage you could also try to make a test case covering every possible branch of your code. The amount of code you execute in a set of tests is referred to as the "code coverage" of the tests.
```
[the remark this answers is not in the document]
Crafty15: Is it common not to test for this behaviour? I was under the impression that overflow should be tested and handled. At least, the examples i've read online do mention it as an edge case.

Crafty15 (commented):
The overflow thing is still a bit hazy to me. Most reference material i've read mentions handling min - 1 and max + 1 (values across boundaries). I suppose we have cases where we don't want to do this. If so, maybe we could state our reasoning.

JosephusPaye:
As to the question of [testing overflow/boundary conditions](https://github.com/NUbots/NUbook/pull/138#discussion_r669205688):

If the thing you're testing could get under/overflowed input as a part of its normal operation, then you want to test that. If it doesn't, then you don't have to test it. Put another way, if you're testing a function that has explicit knowledge of boundary or exceptional conditions (it uses a conditional, try/catch, checks that its inputs are not out of bounds, etc), you want to test on and around those boundaries and conditions.

Essentially, you want your tests to cover as many code paths as possible. Most of the time the boundary conditions you'll be testing are based on some specific numeric value in the code being tested. For example, if you're testing a function that takes an input `x` (or an input that influences `x`) and has `if (x <= 100) { ... }` somewhere in it, you want to test `x = 99`, `x = 100`, and `x = 101` at a minimum.

CMurtagh-LGTM:
> As to the question of [whether or not you should test overflow/underflow/similar boundary conditions](https://github.com/NUbots/NUbook/pull/138#discussion_r669205688):
> 
> If the piece of code you're testing could get under/overflowed input as a part of its normal operation, then you want to test that. If it doesn't, then you don't have to test it. Put another way, if you're testing a function that has explicit knowledge of boundary or exceptional conditions (it uses a conditional, try/catch, checks that its inputs are not out of bounds, etc), you want to test on and around those boundaries and conditions.
> 
> Essentially, you want your tests to cover as many code paths as possible, to verify that everything it attempts to do, it does correctly. Most of the time the boundary conditions you'll be testing will be based on some specific numeric value in the code being tested. For example, if you're testing a function that takes an input `x` (or an input that influences `x`) and has `if (x <= 100) { ... }` somewhere in it, you want to test `x = 99`, `x = 100`, and `x = 101`, at a minimum.

Bounds checking is completely different to [integer overflow](https://en.wikipedia.org/wiki/Integer_overflow). Why are you linking my comment for integer overflow then talking about bounds checking???

Integer overflow is if you have 0111 as a 4 bit integer with two's complement, so this would represent 7. Then if we add one we get 1000, which with two's complement is -8. The c++ standard lists this as [undefined behaviour](https://en.cppreference.com/w/cpp/language/ub)

src/book/03-guides/04-general/04-writing-tests.mdx
ysims: I've been overruled 😆 sounds like I need to git gud

JosephusPaye:
I was responding to the thread your comment started. It would have been more accurate to link to Liam's comment in that thread.

That said, I wasn't talking about bound checking specifically, but the general testing of boundary conditions (of which [arithmetic integer over/underflows](https://www.informit.com/articles/article.aspx?p=686170&seqNum=4) are a subset).

src/book/03-guides/02-tools/05-catch-guide.mdx
```diff
@@ -0,0 +1,65 @@
...
+slug: /guides/tools/catch-guide
+---
+
+[Catch](https://github.com/catchorg/Catch2) is the C++ testing framework which we use for our unit tests. In this guide we will go over the basics of Catch by completing a concrete example. This guide assumes familiarity with C++ and testing concepts. To brush-up on testing, see the [Guide for Writing Tests](/guides/general/writing-tests).
+
+Note that the [Catch docs](https://github.com/catchorg/Catch2/tree/devel/docs) should be the first thing to look at if you're wondering about a specific Catch-ism.
+
+## A Basic Example
+
+Catch test cases have a few components. The most important is the `TEST_CASE(..)` macro, which wraps around groups of associated tests. Inside each `TEST_CASE` scope, you should follow the **AAA** structure, _Arranging_ the data first - both inputs and expected outputs - then _Acting_ by calling the function you're testing, then _Asserting_ that the results match the ground truth. We'll make an example for the following toy utility function:
```
Bidski, line 15: You use `expected output` but `ground truth` elsewhere. You should pick one and stick with it. I think `expected output` is better.

src/book/03-guides/02-tools/05-catch-guide.mdx
Bidski, line 23:
```suggestion
To test it, we'll need pairs of inputs and their expected outputs. Usually, if there's a lot of data, we'll want to keep it separate from the test logic, but for this example we'll keep it local. Also note that for utilities like this one, we would want a much more comprehensive set of test cases.
```

Bidski (dismissed):
Partial review for now

src/book/03-guides/04-general/04-writing-tests.mdx
KipHamiltons: honestly I am torn about this. It adds "needless" complexity, but at the same time I was thinking it should be code that I would actually consider approving. I guess that's maybe a stretch in itself though, because I would not approve a function to add two integers :joy:

src/book/03-guides/04-general/04-writing-tests.mdx
KipHamiltons: excellent point. will do

src/book/03-guides/04-general/04-writing-tests.mdx
KipHamiltons: thinking of answer on the spot, perhaps an allocator which uses address space layout randomisation? I am not sure. I felt like just saying "which generates randomness" wasn't complete enough, but now that you raise it, I am not sure if there are any valid examples like that.

src/book/03-guides/04-general/04-writing-tests.mdx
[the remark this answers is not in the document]
KipHamiltons: Exactly. We don't want to test undefined behaviour - we want to test that our functions avoid the undefined behaviour. So if a function can receive inputs that would cause an underflow, it avoids actually doing it. We can (and should) test that we avoid it.

src/book/03-guides/04-general/04-writing-tests.mdx
KipHamiltons: It is part of philosophy though :thinking: 
I can change the heading if you like

CMurtagh-LGTM: :shrug:

KipHamiltons: I'll change it :+1:

PR #144, opened 2021-07-17
Adds important supplement to Page Not Found page

LachlanCourt:
[Preview](https://deploy-preview-144--nubook.netlify.app/a)
error404


TrentHouliston (commented):
That gif might be a bit big for a 404 1MB isn't so small... maybe cut off the end after the ball stops moving?
Also can the GIF be optimised at all?

LachlanCourt:
It's already pretty shocking quality 😅 I'll run it through an optimiser
I thought the bit where it rolls onto it's back was funny!

BrendanAnnable:
giphy.com uses webp, can we do that? :D

LachlanCourt:
I'll look into it

BrendanAnnable:
For the record, I just meant giphy (a large GIF provider service), actually uses the animated WebP file format instead of GIF. This is because GIF is a terrible file format.

We can still host the file locally in the repo without using giphy as an embed, but just saving it as a `.webp` file, which should be much much much smaller (and higher quality).

LachlanCourt:
It ends up being about the same file size :(

KipHamiltons:
This is brilliant :joy:

ysims (approved):
LGTM 😄

PR #150, opened 2021-07-26
Change team photo

LachlanCourt:
[Preview](https://deploy-preview-150--nubook.netlify.app/team/current-members)


LachlanCourt:
Should I remove the old photo from the source?

ysims:
Yeah remove the old photo. Add in names. And I think we should also get everyone's permission to put it online (although since it was originally going to RoboCup who could use it in any way, it should be fine).

ysims (approved):
Yeah looks good 👍

PR #157, opened 2021-08-13
Upgrade dependencies

JosephusPaye:
[Preview](https://deploy-preview-157--nubook.netlify.app)

Many of our current dependencies are pretty old, and we're getting lots of dependency-related warnings when running `yarn` and `yarn build`.

This PR upgrades most of our dependencies to the latest versions, including Gatsby, Webpack, React, and others. With these come several performance improvements, primarily: 

- the time it takes for `yarn dev` is down substantially, as Gatsby now builds the pages lazily on demand, and not all up front
- the time from file save in editor to updated preview in browser is now significantly shorter

Some of the upgraded packages have had a major version change with backward incompatible changes. I've fixed all the ones I noticed, but there's a possibility of something broken that I missed.

How to review:

- Check all the pages [in the preview](https://deploy-preview-157--nubook.netlify.app), click links, buttons, etc and check if anything is broken
- Clone it locally (as a fresh instance), run `yarn`, then:
  - Run `yarn dev`, make a change on one of the pages, check that it works
  - Run `yarn build`, check that it works
     - If you get this error, ignore - it's expected (it's really only an info/warning, not an error, and silencing it requires a change to an upstream dependency):

       ```
       [BABEL] Note: The code generator has deoptimised the styling of undefined as it exceeds the max of 500KB.
       ```


ysims:
Everything looks fine in the preview.
Should search work when viewing with `yarn dev` or not? (it doesn't) Otherwise everything else looks fine in yarn dev.
I built (yarn build) and was able to open the individual pages by finding them in `public` but idk how to properly view it (clicking on links goes to the wrong places so I think I'm doing it wrong). Can't see any documentation on yarn build in NUbook.

JosephusPaye:
> Everything looks fine in the preview.

👍 

> Should search work when viewing with `yarn dev` or not? (it doesn't) Otherwise everything else looks fine in yarn dev.

It should, but I suspect it might be broken on `master` as well. If that's the case a separate PR to fix it would be better, if not I can fix it here.

> I built (yarn build) and was able to open the individual pages by finding them in `public` but idk how to properly view it (clicking on links goes to the wrong places so I think I'm doing it wrong). Can't see any documentation on yarn build in NUbook.

`yarn build` is primarily meant for use when the site is deployed (e.g. by Netlify). Though it can be useful for debugging the build system locally, so it's probably worth documenting.

After it's built, you need a web server to view the files in the `public/` folder correctly. You can use this:

```
npx live-server public/
```

ysims (approved):
`main` is taking forever to generate the site with `yarn dev` so as far as I can see this PR is a large improvement on `main`. Here's an approval.

PR #164, opened 2021-09-12
Updates clone command on the Webots set up page

ysims:
Current line gave me an error. It doesn't need the recurse submodules stuff anyway because that has to be done in a separate command later (`git submodule update --init`) because we're doing weird things with submodules inside a subtree.


KipHamiltons:
Shouldn't we be running `git submodule update --init --recursive` for that call? Doesn't webots (now a submodule of ours) have submodules itself, which implies recursive submodule init is required?

ysims:
> Shouldn't we be running `git submodule update --init --recursive` for that call? Doesn't webots (now a submodule of ours) have submodules itself, which implies recursive submodule init is required? Did this work without that extra flag?

Yeah, it worked. I deleted my `NUWebots` folder and rebuilt everything this morning because of some issue with switching to an old branch.

KipHamiltons (approved):
I am not sure why it works, but sure :shrug:

PR #172, opened 2021-09-15
Adds a system diagram in GraphViz to the Overview Page

ysims:
Decided to remove transitive edges and not label the edges, to make it more viewable. For the most part it's ok, but there are some things where this more so is what our system _should be_ rather than what it _is right now_. Ie we don't actually use localisation atm since it's broken. 

https://deploy-preview-172--nubook.netlify.app/system/foundations/overview#system-diagram


CMurtagh-LGTM (approved):

CMurtagh-LGTM (approved):

JosephusPaye (approved):
LGTM. Now I'm just sad that the diagram at the top of the page doesn't have the auto light/dark mode switching that the graphviz ones have :P

ysims:
I should probably convert that diagram lol, or remove it :thinking:

PR #193, opened 2021-10-16
Updates the current members list (October 2021)

ysims:
We have new Junior Members! I've added them to the list.
Also minor changes to groups and positions of existing members.
https://deploy-preview-193--nubook.netlify.app/team/current-members


miikyla:
Did only 20 accept? I swear it was 21

Issue #186, opened 2021-10-05
Add a "Last updated" and "edit" element to NUbook pages to make it clear if things have gotten out of date


LachlanCourt:
See slight discussion on #175

JosephusPaye:
Mobile might look something like:

![Screenshot showing article meta data section on mobile](https://user-images.githubusercontent.com/5924865/136780988-05c8395c-523b-4bb4-854a-84eda211dd66.png)

PR #197, opened 2021-11-25
Add Contributors component

lach-j:
Addresses [#186](https://github.com/NUbots/NUbook/issues/186)

Adds Last Updated section to pages as part of the Contributors component to show when the page was last updated.

[Preview](https://deploy-preview-197--nubook.netlify.app/guides/main/getting-started)

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -37,6 +37,18 @@ module.exports = {
         path: `${__dirname}/src/pages`,
       },
     },
+    {
+      // Adds configuration for querying GitHub GraphQl API
+      resolve: 'gatsby-source-graphql',
+      options: {
+        typeName: 'GitHub',
+        fieldName: 'github',
+        url: 'https://api.github.com/graphql',
+        headers: {
+          Authorization: process.env.GITHUB_TOKEN,
+        },
+      },
+    },
     'gatsby-plugin-image',
     // Exposes helper functions for processing images with the
     // sharp package from npm. Used by other plugins.
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -45,7 +45,7 @@ module.exports = {
         fieldName: 'github',
         url: 'https://api.github.com/graphql',
         headers: {
-          Authorization: process.env.GITHUB_TOKEN,
+          Authorization: `Bearer ${process.env.GITHUB_TOKEN}`,
         },
       },
     },
--- a/src/components/contributions/contributions.jsx
+++ b/src/components/contributions/contributions.jsx
@@ -2,20 +2,18 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import EditIcon from './edit-icon.svg'
 
-const Contributions = ({ data }) => {
-  const commit = data.github.repository.object.history.nodes[0]
-  const file = data.github.repository.object.file
-  const lastUpdated = new Date(commit.author.date).toLocaleString('en-AU', {
+const Contributions = ({ date, url, mdxPath }) => {
+  const lastUpdated = new Date(date).toLocaleString('en-AU', {
     dateStyle: 'long',
   })
-  const editLink = `https://github.com/NUbots/NUbook/edit/main/${file.path}`
+  const editLink = `https://github.com/NUbots/NUbook/edit/main/${mdxPath}`
   return (
     <div className='text-sm text-secondary dark:text-secondary-inverted mb-6 border-t-2 border-b-2 border-gray-200 dark:border-gray-800 py-3 flex'>
       <div className='flex-grow'></div>
       <div className='pr-3 capitalize'>
         <span className='hidden lg:inline'>Last</span> updated{' '}
         <a
-          href={commit.url}
+          href={url}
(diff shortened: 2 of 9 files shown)


src/components/page-template.jsx
```diff
@@ -15,6 +15,23 @@ const PageTemplate = (props) => {
 
 PageTemplate.propTypes = {
   data: PropTypes.shape({
```
LachlanCourt, line 17: Because we are only pulling out the date and the url, I think it will be neater to have this interpretted in PageTemplate function above, just before the return, and just pass through the date and url instead of the entire graphql response. I'm putting this here because I can't comment on the actual line as it wasn't changed in this PR 😆

src/components/contributions/contributions.jsx
```diff
@@ -0,0 +1,62 @@
...
+          <span className='pl-1 hidden lg:block text-nubots-700 dark:text-nubots-500'>
+            Edit
+          </span>
+        </a>
+      </div>
+    </div>
+  )
+}
+
+Contributions.propTypes = {
```
LachlanCourt, line 40: See comment below about passing through less props, I think just the latest commit and the url need to be passed through here rather than the entire graphql response. This will also mean less updates if the graphql API changes in the future. This also means that the prop names can be more descriptive than `data`

src/components/contributions/contributions.jsx
```diff
@@ -0,0 +1,62 @@
...
+    <div className='text-sm text-secondary dark:text-secondary-inverted mb-6 border-t-2 border-b-2 border-gray-200 dark:border-gray-800 py-3 flex'>
+      <div className='flex-grow'></div>
+      <div className='pr-3 capitalize'>
+        <a
+          href={commit.url}
+          target='_blank'
+          className='lg:normal-case'
+          rel='noreferrer'
+        >
+          <span className='hidden lg:inline'>Last</span> updated{' '}
```
LachlanCourt, line 22: Any reason why `updated` is outside of the span?

src/components/contributions/contributions.jsx
```diff
@@ -0,0 +1,62 @@
...
+  const file = data.github.repository.object.file
+  const lastUpdated = new Date(commit.author.date).toLocaleString('en-AU', {
+    dateStyle: 'long',
+  })
+  const editLink = `https://github.com/NUbots/NUbook/edit/main/${file.path}`
+  return (
+    <div className='text-sm text-secondary dark:text-secondary-inverted mb-6 border-t-2 border-b-2 border-gray-200 dark:border-gray-800 py-3 flex'>
+      <div className='flex-grow'></div>
+      <div className='pr-3 capitalize'>
+        <a
```
LachlanCourt, line 16: Because only the date is coloured, I think it's best to move this hyperlink so that it only wraps the date, not the words `Last updated`

LachlanCourt (changes requested):
This looks awesome! Really nice work. Just a few semantic changes to tidy it up a little.

LachlanCourt:
I added a preview of the deployment too so you can see the live site. This will update every time you push changes, same link 🙂

src/components/contributions/contributions.jsx
lach-j: This was done so that on mobile 'Last' is hidden and only 'Updated DATE' is shown.

src/components/contributions/contributions.jsx
lach-j: Ok sounds good. This is also the case for the edit icon and text however the icon is grey and 'edit' is not shown on mobile. Would it make sense to also change these for consistency? Maybe make the icon coloured on mobile?

LachlanCourt: I think the functionality of the edit is fine as is, we have other logos which are grey and also links in NUbook but I think semantically just having the date as the actual hyperlink for the updated will be more intuitive

src/components/contributions/contributions.jsx
LachlanCourt: Gotcha. Nice!

gatsby-config.js
```diff
@@ -37,6 +37,18 @@ module.exports = {
...
     },
+    {
+      // Adds configuration for querying GitHub GraphQl API
+      resolve: 'gatsby-source-graphql',
+      options: {
+        typeName: 'GitHub',
+        fieldName: 'github',
+        url: 'https://api.github.com/graphql',
+        headers: {
+          Authorization: `Bearer ${process.env.GITHUB_TOKEN}`,
```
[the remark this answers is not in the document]
LachlanCourt: Got this running using my own access token but we should move this to a centralised account at some point @ysims

src/components/page-template.jsx
JosephusPaye: +1

LachlanCourt (approved):
I think this looks great!
My only concern is potential rate limits of the graphql api but I find it hard to believe that we will reach those. [Rate limits here](https://docs.github.com/en/graphql/overview/resource-limitations)

PR #201, opened 2021-12-03
Increases the margin in maths blocks

ysims:
https://deploy-preview-201--nubook.netlify.app/kitchen-sink/math
https://deploy-preview-201--nubook.netlify.app/system/subsystems/vision#ball-detector

[The vision page](https://nubook.nubots.net/system/subsystems/vision#ball-detector) has maths equations in a list. There doesn't appear to be any margin or padding, making it very squashed and visually unappealing/harder to read. 

This PR aims to increase this margin. I added to the margin that was already in the markdown CSS file and it doesn't appear to affect the margin of normal maths blocks, but does increase it for other components like the list and grid. I increased the margin to the point just before it started to affect the normal blocks. 

Before:
image
After: 
image

This may not be a great way to do this - I feel like it's probably better to somehow do it specifically for the list scenario than globally?


JosephusPaye (commented):
Looks good! The extra spacing is much better for readability. I'd just add headings to match the other ones on the kitchen sink:

- Inline
- Block
- Large block
- In a list
- In collapsible

JosephusPaye:
> This may not be a great way to do this - I feel like it's probably better to somehow do it specifically for the list scenario than globally?

Do the non-list ones have good spacing when there are multiple large ones next to each other? If not then we should probably apply it globally to fix those too.

ysims:
> It looks like we set the spacing for normal blocks together with the spacing for everything else at the top level of the markdown content and inside collapsibles ([see here](https://github.com/NUbots/NUbook/blob/9ae0bcb3e578284db743f0a543a956b25408628c/src/components/markdown/markdown.module.css#L18)). This means we only need to set the spacing for math blocks in lists and grids.

Ah yeah haha I just was experimenting and it was just the lists and grid. The grid is a bit close but changing it to the same as the list is a bit far apart. 

> For grids, we could add something similar to the above in https://github.com/NUbots/NUbook/blob/main/src/components/markdown/grid/grid.module.css, but I think it might be more useful to add a gap property, in addition to the existing columns on the <Grid> element.

Yeah I think I'd agree with that. The same spacing as the lists seems a bit much for the grid, but it depends on context maybe.

JosephusPaye:
Made a PR to add `gap` property to `<Grid>`: https://github.com/NUbots/NUbook/pull/203

LachlanCourt (approved):
I think this is fine
There's still some weirdness regarding the block quote and the table but I think overall it's a lot better than it was

PR #202, opened 2021-12-04
Added Webots windows error message

JamieTsang19:
A potential problem when setting up Webots on Windows was mentioned, but the error given was not stated. This PR adds that error message.


LachlanCourt (changes requested):
If we're not using the image anymore we may as well delete it from the repo

LachlanCourt:
You might need to run `yarn format` locally to fix this one up :) sometimes NUbook commits done on github are hard to get the formatting right
Do you want to remove the image while you're at it as we won't be using it anymore @JamieTsang19 ?

LachlanCourt (changes requested):
You might need to run `yarn format` locally to fix this one up :) sometimes NUbook commits done on github are hard to get the formatting right
Do you want to remove the image while you're at it as we won't be using it anymore @JamieTsang19 ?

ysims (approved):

Issue #192, opened 2021-10-11
Add contributor list to NUbook pages


LachlanCourt:
Will need consideration on how to display this on mobile.
Should be completed after issue [#186](https://github.com/NUbots/NUbook/issues/186)

![image](https://user-images.githubusercontent.com/80299053/136778058-892b4fdb-1248-406d-a8c0-a606ff076a66.png)

PR #204, opened 2021-12-07
Add contributor list to contributions section

lach-j:
Addresses #192 

- Adds list of page contributors to the contributions/last updated section.

The name tool-tips currently use the `title` attribute as I had trouble implementing the tool-tips shown in the [mock-up](https://github.com/NUbots/NUbook/issues/192#issue-1022549762).

[Preview](https://deploy-preview-204--nubook.netlify.app/system/foundations/build-system)


ysims:
Looks good. Seems that it can be a bit misleading though, there are a fair few pages that don't list the main writer because the file name was changed.

lach-j:
Yeah, I'm not sure if that can be fixed because of the way git treats renaming :confused:. 

The way I have it set up is its ordered by the most recent commit author so that's also a bit misleading. I could order it based on number of additions and deletions per author if that would better represent it.

LachlanCourt (approved):
😍

PR #183, opened 2021-10-02
Adds case conventions

ysims:
Addresses #176. 

I've repurposed the code doc page into a code conventions page. Lemme know if you think it should go somewhere else.

- Classes are PascalCase
- Functions are snake_case
- Variables are snake_case
- yaml fields are snake_case
- Constants are SCREAMING_SNAKE_CASE

PR #206, opened 2022-01-15
Fix contributors component accessing undefined properties on new pages

lach-j:
Ensures that the contributor component only renders when a commit is returned by the GitHub API. 

This prevents the contributor component from accessing undefined properties on new or renamed pages that don't exist on the main branch (see [discussion](https://github.com/NUbots/NUbook/pull/183#issuecomment-1013558037)).


LachlanCourt (approved):
Makes sense

JosephusPaye:
This is good for now to fix the current issue, but I think we should consider changing the way we're getting the data for the contributors list. See this proposal: #207.

Issue #207, opened 2022-01-15
Proposal: Change the way we get data for the contributors list


JosephusPaye:
I think we should switch away from the current way we get the data to show the list of contributors to a page (using the git history via the GitHub API) to instead use a list of authors at the top of the file in the frontmatter:

```yaml
---
section: The NUbots Team
chapter: Introduction
title: Introduction to NUbots
description: Learn about what we do, key people, and where to find the lab.
slug: /team/introduction
authors:
  - Ysobel Sims (@ysims)
  - Josephus Paye II (@JosephusPaye)
  - Lachlan Court (@LachlanCourt)
---
```

`authors` will be a list of contributors to the page, with their name followed by GitHub username in parenthesis. Each new person that makes a change to the page simply adds their name and username to the end of the list.

We can then use the author's GithHub username to link to their profile, as well as get their avatar. Given a username, the GitHub profile is at `https://github.com/@username` and the avatar is at `https://github.com/@username.png`.

For the last updated date and link, we can run a git command locally for each page during the build to get the latest commit hash and date. E.g. running the following...

```bash
git log -n1 --pretty='format:%H %cd' --date=format:'%Y-%m-%d' -- src/book/01-team/01-introduction/01-introduction-to-n
ubots.mdx
```

... outputs the commit hash and date:

```
8877c451459e18e6295a7f90d9207f67ef81ce94 2021-07-21
```

The hash can be used to link to the commit: `https://github.com/NUbots/NUbook/commit/<hash_here>`

This approach has many benefits over the current one:

- Resilient to file changes: renaming a file may sometimes loose the edit history, this avoids that
- Resilient to things like auto formatting: you won't get added to the list of contributors if we change the code style and you reformat all files
- Simpler to implement: we don't need to use the GitHub API to fetch the list of contributors at build time

One (albeit minor) downside: if we do this, we'll need a PR that goes through and fills in the details for the current pages.

lach-j:
Seems like a much more reliable and flexible way of doing things and would eliminate all the GitHub issues too.

It doesn't look quite as neat but it may make things easier to organise the authors like this so we don't have to try to split the name and usernames later.
```yaml
---
section: The NUbots Team
chapter: Introduction
title: Introduction to NUbots
description: Learn about what we do, key people, and where to find the lab.
slug: /team/introduction
authors:
  - name: Ysobel Sims
    login: ysims
  - name: Josephus Paye II
    login: JosephusPaye
  - name: Lachlan Court
[...]

PR #209, opened 2022-01-22
Update the contributors list to source data locally

JosephusPaye:
Implements the proposal in #207. Also makes minor simplifications to the display of contributors.

I was gonna to work on #69 but merging `main` brought up errors about the GH API token for the contributions so I switched to doing this instead.

I've added the data for the Introduction to NUbots page, and we can fill out the rest in another PR.

Preview: [Introduction to NUbots](https://deploy-preview-209--nubook.netlify.app/team/introduction)


lach-j (approved):
Looks great :smile:

Only small issue I can see is that once the authors get added to the existing pages, the dates will reflect that commit instead of when content was added. Shouldn't really matter too much though and will fix itself eventually :shrug:

src/images/nubots-icon.svg
```diff
@@ -0,0 +1 @@
...-.218-.829-.287-1.336-.324-2.024-.053-.696 1.423-1.058 1.718-2.513 2.227-.622.217-1.436.466-1.81.556-.24.057-.5.098-.758.123a18 18 0 00-.1.642 9.161 9.161 0 003.455-.979c1.12-.555 1.617-.668 2.147-.488l.103.034a14.709 14.709 0 01-8.327 2.667l-.133.002c-2.34 0-4.56-.551-6.542-1.534v-.007c-.018-.664-.416-2.11-.735-2.674-.374-.659-1.336-1.788-2.911-3.418a21.06 21.06 0 00-.938-.927c.094.147.186.294.27.433.333.555.836 1.29 1.118 1.63.983 1.19 1.07 1.42 1.07 2.817v.863a15.731 15.731 0 01-5.516-7.275l.473-.016-.397-.173-.172-.075a16.139 16.139 0 01-.912-5.365 16.214 16.214 0 01.961-5.504A15.733 15.733 0 017.954 3.53a14.726 14.726 0 018.883-2.28l.256.16c.338.213.803.46 1.033.55.09.033.242.092.429.164-.14-.181-.27-.366-.379-.543l-.124-.202a14.868 14.868 0 017.167 3.17c.51.413.995.86 1.45 1.336a15.82 15.82 0 013.154 4.82 4.57 4.57 0 01-.116.374c-.176.508-.173.534.125 1.739.362 1.457.462 3.226.26 4.576-.062.404-.09.84-.075 1.063.393-2.272.55-2.955.812-3.558l.086-.199a16.244 16.244 0 01-.066 4.912 15.968 15.968 0 01-1.55 4.687zM15.877.772C7.262.772.253 8.024.253 16.94c0 8.914 7.01 16.166 15.624 16.166 8.616 0 15.625-7.252 15.625-16.166 0-8.915-7.01-16.167-15.625-16.167z" fill="#F9A50D"/></svg>
```
ysims, line 1: How come this has been added? I don't see a nubots icon in the contributors/last edited bar?

ysims (commented):
You've got a `yarn.lock` file in your changes.

src/images/nubots-icon.svg
JosephusPaye: It's used for authors without a GitHub profile, as the default avatar. E.g. if someone like Stephan (assuming he doesn't have a GH account) were to add to a page they can be credited with just their name, and the NUbots icon will be used, without linking to GitHub.

JosephusPaye:
The `yarn.lock` change is expected: it happens whenever a dependency is added, removed, or upgraded. In this case the change is due to removing the `gatsby-source-graphql` package, which was used for talking to the GitHub API for commits and contributors.

As to why we even have lock files in the first place, [see this article](https://snyk.io/blog/what-is-package-lock-json/).

src/images/nubots-icon.svg
ysims: Ah, ok.

ysims (commented):
Can't really comment on the js but I can't see any issues. 
I'm gonna agree with the approach since even though it is more manual, it is much more correct. I don't really want to be credited for a page I only reformatted or renamed :sweat_smile:.

ysims (approved):
:+1:

PR #210, opened 2022-02-01
Ensure files in `build-plugins/` and `static/` are formatted

JosephusPaye:
Currently files in `build-plugins/` and `static/` are not checked for formatting with `yarn format:check`. This PR remedies that, and fixes a formatting issue in two of the files.


ysims (commented):
Should the static folder also be added? If it isn't already.

JosephusPaye:
Yep, good catch!

ysims (approved):
LGTM :+1:

PR #221, opened 2022-06-01
Fix bug where search bar stops working after clicking NUbots logo

LachlanCourt:
[preview](https://deploy-preview-221--nubook.netlify.app/)

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,8 +61,8 @@
     "dev": "gatsby develop",
     "clean": "rimraf .cache/ public/",
     "lint": "eslint \"src/**/*.{jsx,js}\" \"*.js\" && stylelint \"src/**/*.css\"",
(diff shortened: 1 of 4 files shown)


JosephusPaye:
Looks good 👍 

At this point we may as well upgrade to v3 tho? 😅 https://docsearch.algolia.com/docs/DocSearch-v3

Comes with built-in React components we can `yarn add` and use pretty easily:

```jsx
import { DocSearch } from '@docsearch/react';

import '@docsearch/css';

function App() {
  return (
    <DocSearch
      appId="YOUR_APP_ID"
      indexName="YOUR_INDEX_NAME"
      apiKey="YOUR_SEARCH_API_KEY"
    />
  );
}

export default App;
```

Might just need some few styling tweaks to match what we did here: https://github.com/NUbots/NUbook/blob/df4c670b3d9e612c0e97588fd3a956496490539f/src/styles/global.css#L101-L192

ysims (approved):
LGTM. Updating Algolia can be done in a different PR.

PR #223, opened 2022-06-03
Update config info to include platform-specific and alert about docker/webots hostname.

ysims:
Updates NUbook for NUbots/NUbots#833.
The config system now includes platform-specific config folders.
In Docker, if the binary being run includes the word `webots`, it will have `webots` as the hostname.


src/book/02-system/01-foundations/06-config-script.mdx
```diff
@@ -40,16 +43,19 @@ However, it makes no sense for the default configuration file to contain the
...
 
 1. The default configuration file `config/ModuleName.yaml` is loaded. This forms the base configuration for the module
-1. A check is made to see if `config/RobotName/ModuleName.yaml` exists. If it does, it is loaded and all values found therein override the equivalent values in the base configuration
-1. A check is made to see if `config/RoleName/ModuleName.yaml` exists. If it does, it is loaded and all values found therein override the equivalent values in the base configuration
-1. If any of these 3 configuration files are modified then 1-3 are run again
+2. A check is made to see if `config/PlatformName/ModuleName.yaml` exists. If it does, it is loaded and all values found therein override the equivalent values in the base configuration
+3. A check is made to see if `config/RobotName/ModuleName.yaml` exists. If it does, it is loaded and all values found therein override the equivalent values in the base configuration
+4. A check is made to see if `config/RoleName/ModuleName.yaml` exists. If it does, it is loaded and all values found therein override the equivalent values in the base configuration
```
LachlanCourt, line 57:
```suggestion
2. A check is made to see if `config/<PlatformName>/<ModuleName>.yaml` exists. If it does, it is loaded and all values found therein override the equivalent values in the base configuration
3. A check is made to see if `config/<RobotName>/<ModuleName>.yaml` exists. If it does, it is loaded and all values found therein override the equivalent values in the base configuration
4. A check is made to see if `config/<RoleName>/<ModuleName>.yaml` exists. If it does, it is loaded and all values found therein override the equivalent values in the base configuration
```

src/book/02-system/01-foundations/06-config-script.mdx
```diff
@@ -40,16 +43,19 @@ However, it makes no sense for the default configuration file to contain the
...
+5. If any of these 3 configuration files are modified then 1-4 are run again
```
LachlanCourt, line 58:
```suggestion
5. If any of these 3 configuration files are modified then steps 1-4 are run again
```

LachlanCourt (changes requested):
Don't forget to add yourself as a contributor in the metadata

src/book/02-system/01-foundations/06-config-script.mdx
ysims: I feel like this might be too messy with the number of things, plus we'd have to do it below for scripts. I feel like this is more of an example, following the example above.

LachlanCourt (approved):

PR #225, opened 2022-06-07
Fix intermittent production build issues

LachlanCourt:
[preview](https://deploy-preview-225--nubook.netlify.app/)


src/components/sidebar/sidebar.jsx
```diff
@@ -40,7 +40,9 @@ const Sidebar = ({ menu, currentSection, wrapperRef }) => {
 
   // Scroll to reveal the active link on mount
   useEffect(() => {
-    const activeLink = sidebarRef.current.querySelector(`.${style.linkActive}`)
+    const activeLink = sidebarRef.current.querySelector(
+      `.${style.linkActive.replace('+', '\\+')}`
```
LachlanCourt, line 44: `+` is not a valid classname :angry:

JosephusPaye: Does the active link classname include it?

LachlanCourt: Yup

ysims (commented):
Went through the pages and looks good, cant see any issues.

src/components/sidebar/sidebar.jsx
ysims: Does it the active link class name have + for a reason, or can it be changed rather than doing this here?

LachlanCourt: It gets generated by Gatsby. It wouldn't actually be a problem as a class name usually except it's not officially a valid character and so the query selector doesn't work. The `+` is a [css selector symbol](https://www.w3schools.com/cssref/css_selectors.asp) and so if it's in the classname then it is interpreting it as that instead so we need to escape it

src/components/sidebar/sidebar.jsx
```diff
@@ -29,9 +31,8 @@ const getLinkProps = ({ isCurrent }) => {
   return {
     className: `${
       style.link
-    } text-primary-muted dark:text-primary-muted-inverted ${
-      isCurrent ? style.linkActive : ''
-    }`,
+    } text-primary-muted dark:text-primary-muted-inverted 
+    ${isCurrent ? `${style.linkActive} ${ACTIVE_LINK_SELECTOR}` : ''}`,
```
LachlanCourt, line 35: This is another solution @JosephusPaye

src/components/sidebar/sidebar.jsx
LachlanCourt: Supposedly there's a way to override the active link classname however I tried it and it did not work, I think because we are passing in a class name manually

src/components/sidebar/sidebar.jsx
JosephusPaye: It's so weird that the CSS module import is generating a class name with `+` in it.

Anyway, I agree that using a different selector for the active link is the way to go. In cases where I'm adding a selector solely for JS purposes, I like to use a data attribute ([some of the reasons covered here](https://pjmcdermott92.medium.com/why-i-use-custom-data-attributes-for-selecting-elements-in-javascript-58166fe074a8)).

With that approach, `getLinkProps()` will look like:

```js
const getLinkProps = ({ isCurrent }) => {
  return {
    'className': `${
      style.link
    } text-primary-muted dark:text-primary-muted-inverted
    ${isCurrent ? style.linkActive : ''}`,
    'data-active-link': isCurrent ? true : null,
  }
}
```

And the selector will look like:

```js
const activeLink = sidebarRef.current.querySelector('[data-active-link]')
```

LachlanCourt: Makes sense! Thanks :)

src/components/sidebar/sidebar.jsx
```diff
@@ -27,11 +27,12 @@ function elementInView(element, container) {
 
 const getLinkProps = ({ isCurrent }) => {
   return {
-    className: `${
+    'className': `${
       style.link
-    } text-primary-muted dark:text-primary-muted-inverted ${
+    } text-primary-muted dark:text-primary-muted-inverted  ${
```
JosephusPaye, line 32: Minor nit

```suggestion
    } text-primary-muted dark:text-primary-muted-inverted ${
```

JosephusPaye (approved):
LGTM! 🎉

src/components/sidebar/sidebar.jsx
arabjavad36: #

PR #228, opened 2022-06-09
Add instructions for changing wifi networks

mrwerdo:
There's a nice way of changing wifi networks without needing to modify any settings. It's done using the `wpa_cli` program which manages the associated wifi network:

`wpa_cli select_network 0` selects the 0th configured network from `/etc/wpa_supplicant/wpa_supplicant-wlp58s0.conf`.

The advantage of using this approach is 1) no root access is needed, 2) no permanent changes to the network files, and 3) only two commands.

Let me know what you people think. There's another question about whether a test network could be configured for the robots during the flashing process. Do you think that would be a good idea?

### DIFF
--- a/src/book/03-guides/03-hardware/02-flashing.mdx
+++ b/src/book/03-guides/03-hardware/02-flashing.mdx
@@ -208,3 +208,5 @@ In summary, the three commands that you need to know are:
 3. `wpa_cli status`
 
 No file modifications or root access required.
+
+It's recommended to not have the robot connected to the ethernet network. This code assumes that the IP address of the robot, and in general, the network topology is similar enough between the two networks.
--- a/src/book/03-guides/03-hardware/02-flashing.mdx
+++ b/src/book/03-guides/03-hardware/02-flashing.mdx
@@ -165,27 +165,35 @@ The rest of these instructions assume a network using WPA2 (like the network use
    - Wait a handful of seconds and run `ip addr` to check that the WiFi interface has an IP address
 
 
-## General Network Tips
+## Switching Wifi Networks
 
-Suppose you need to switch wifi networks on the robot. There are two options: 
-1. Modifying a local file (not so good)
-2. Changing which network the robot is associated with (better)
-The second option is better because it does not persist after reboot, which means that any changes are not permenant.
+Switching wifi networks can be accomplished with the following two commands:
 
-The tool you need to use is wpa_cli:
-
-```bash
-nubots@nugus3 ~ ❯❯❯ wpa_cli list_networks
+```
+$ wpa_cli list_networks
 Selected interface 'wlp58s0'
 network id / ssid / bssid / flags
 0	epsilon-z	any	[CURRENT]
 1	epsilon-x	any	[DISABLED]
 2	NETGEAR89	any	[DISABLED]
-nubots@nugus3 ~ ❯❯❯ wpa_cli select_network 2; wpa_cli status
+
+$ wpa_cli select_network 2
 Selected interface 'wlp58s0'
 client_loop: send disconnect: Broken pipe
-# ... different connection.
-nubots@nugus3 ~ ❯❯❯ wpa_cli status
+```
+
+The number `N` provided to `wpa_cli select_network <N>` is the network id given by `wpa_cli list_networks`. To switch to `epsilon-z` one would execute `wpa_cli select_network 0`.
+The example above was executed through a connection over the wifi interface, hence why the pipe broke.
+
+This approach is better than modifying the files because:
+1. Modifying files is permenent where as selecting a different network is not.
+2. Root permissions are not required to change networks.
+3. There's no need to reload the wpa_supplicant configuration.
+
+
+Finally, to check which network the robot is associated with, run:
+```bash
+$ wpa_cli status
 Selected interface 'wlp58s0'
 bssid=14:59:c0:d2:07:29
 freq=2472
@@ -207,6 +215,5 @@ In summary, the three commands that you need to know are:
 2. `wpa_cli select_network <N>`
 3. `wpa_cli status`
 
-No file modifications or root access required.
-
-It's recommended to not have the robot connected to the ethernet network. This code assumes that the IP address of the robot, and in general, the network topology is similar enough between the two networks.
+This technique assumes that the network has been configured in `/etc/wpa_supplicant/wpa_supplicant-wlp58s0.conf`. See above for associating with different wifi networks.
+The bond network configuration on the robot means that a connected ethernet connection will take priority. This would mean packets received over the wifi would be responded to over the ethernet. Given that the ip addresses, gateway and subnets are the same between networks, it probably won't be obvious that packets are being routed incorrectly. (Or it might just work, who knows?).
--- a/src/book/03-guides/03-hardware/02-flashing.mdx
+++ b/src/book/03-guides/03-hardware/02-flashing.mdx
@@ -163,3 +163,48 @@ The rest of these instructions assume a network using WPA2 (like the network use
      ```
    - Be sure to replace `<WiFi Interface>` with the name of the interface found earlier.
    - Wait a handful of seconds and run `ip addr` to check that the WiFi interface has an IP address
+
+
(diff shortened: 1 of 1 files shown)


src/book/03-guides/03-hardware/02-flashing.mdx
```diff
@@ -163,3 +163,57 @@ The rest of these instructions assume a network using WPA2 (like the network use
...
+$ wpa_cli list_networks
+Selected interface 'wlp58s0'
+network id / ssid / bssid / flags
+0	epsilon-z	any	[CURRENT]
+1	epsilon-x	any	[DISABLED]
+2	NETGEAR89	any	[DISABLED]
+
+$ wpa_cli select_network 2
+Selected interface 'wlp58s0'
+client_loop: send disconnect: Broken pipe
```
CMurtagh-LGTM, line 182:
```suggestion
# wpa_cli list_networks
Selected interface 'wlp58s0'
network id / ssid / bssid / flags
0	epsilon-z	any	[CURRENT]
1	epsilon-x	any	[DISABLED]
2	NETGEAR89	any	[DISABLED]

# wpa_cli select_network 2
Selected interface 'wlp58s0'
client_loop: send disconnect: Broken pipe
```
Does this need to be run as root, or is this unnecessary on the robots?
It needs root on my systems and also the arch wiki documents that it needs root.

src/book/03-guides/03-hardware/02-flashing.mdx
```diff
@@ -163,3 +163,57 @@ The rest of these instructions assume a network using WPA2 (like the network use
...
+address=76:0f:5d:3d:ce:db
+uuid=f96d74c5-bf3a-5aa7-ab11-9f3892fb8d25
+```
+
+In summary, the three commands that you need to know are:
+1. `wpa_cli list_networks`
+2. `wpa_cli select_network <N>`
+3. `wpa_cli status`
+
+This technique assumes that the network has been configured in `/etc/wpa_supplicant/wpa_supplicant-wlp58s0.conf`. See above for associating with different wifi networks.
```
CMurtagh-LGTM, line 218:
```suggestion
This technique assumes that the network has been configured in `/etc/wpa_supplicant/wpa_supplicant-wlpXXsX.conf`. See above for associating with different wifi networks.
```
These numbers can change, however unlikely.

src/book/03-guides/03-hardware/02-flashing.mdx
```diff
@@ -163,3 +163,57 @@ The rest of these instructions assume a network using WPA2 (like the network use
...
+
+This approach is better than modifying the files because:
+1. Modifying files is permenent where as selecting a different network is not.
+2. Root permissions are not required to change networks.
+3. There's no need to reload the wpa_supplicant configuration.
+
+
+Finally, to check which network the robot is associated with, run:
+```bash
+$ wpa_cli status
```
CMurtagh-LGTM, line 196:
```suggestion
# wpa_cli status
```
See above

src/book/03-guides/03-hardware/02-flashing.mdx
```diff
@@ -163,3 +163,57 @@ The rest of these instructions assume a network using WPA2 (like the network use
...
+client_loop: send disconnect: Broken pipe
+```
+
+The number `N` provided to `wpa_cli select_network <N>` is the network id given by `wpa_cli list_networks`. To switch to `epsilon-z` one would execute `wpa_cli select_network 0`.
+The example above was executed through a connection over the wifi interface, hence why the pipe broke.
+
+This approach is better than modifying the files because:
+1. Modifying files is permenent where as selecting a different network is not.
+2. Root permissions are not required to change networks.
+3. There's no need to reload the wpa_supplicant configuration.
```
CMurtagh-LGTM, line 191:
```suggestion
2. There's no need to reload the wpa_supplicant configuration.
```
Above

src/book/03-guides/03-hardware/02-flashing.mdx
mrwerdo: Yeah, I guess I was trying to point out where the networks in `wpa_cli list_networks` came from. I'm not sure of a clear way of writing that though. It's sort of related to the instructions earlier in the page.

Bidski:
While this is cool and all, you still need to configure the network for `wpa_cli` to allow you to select it, and outside of the lab (where there is only one network you need to care about) the only other networks you would need to switch to are the robocup fields which you would have to setup at robocup (and will change every year). So I am not really sure why this is needed?

src/book/03-guides/03-hardware/02-flashing.mdx
CMurtagh-LGTM: Looks like [here](https://github.com/NUbots/NUbots/blob/d4d23617295fc66aac75864602fd08053f812614/doc/ArchInstall/arch-chroot_install.sh#L139) we set wheel to be able to change settings.

ysims:
I've had a brief look at this and here's some things
- I think this should live in a new RoboCup guide. The guide would contain a lot of different things about RoboCup. One of them would be about running a test game in the lab, with this information (although my point below will contradict this). Another thing on that page should be how to set up a new network on the robot for RoboCup, which should be quite generic since we don't know the details until we are there. 
- We should investigate why we need a separate router. From what I understand, we have latency on the Asus router but not on Netgear with GameController and the robots. It could just be normal wifi network traffic and the proper solution is to use a different router, but we should try to figure out if we can use the Asus router and why the packets are getting stuck for so long at the router. 
- Is this set up on all of the robots now? If not, can you set it up on all of them? If we don't sort out the problems with the Asus router, we'll want to easily be able to switch the robots to the Netgear router.

ysims:
@Bidski my comment above might explain it, but we do have two networks if we use the Netgear router to make an isolated game network. Ideally though we fix the real problem, if it's fixable.

Bidski:
> Ideally though we fix the real problem, if it's fixable.

This should always be the solution. Band-aid solutions are never particularly effective and they never actually solve anything. 

Having said that, having a guide detailing how to properly setup and configure the wifi network on the robot is not a bad idea. If you are linking to anything external, these two should probably be it [Arch Linux systemd-networkd](https://wiki.archlinux.org/title/systemd-networkd) and [Arch Linux Network Configuration](https://wiki.archlinux.org/title/Network_configuration)

src/book/03-guides/03-hardware/02-flashing.mdx
```diff
@@ -163,3 +163,57 @@ The rest of these instructions assume a network using WPA2 (like the network use
...
+2	NETGEAR89	any	[DISABLED]
+
+$ wpa_cli select_network 2
+Selected interface 'wlp58s0'
+client_loop: send disconnect: Broken pipe
+```
+
+The number `N` provided to `wpa_cli select_network <N>` is the network id given by `wpa_cli list_networks`. To switch to `epsilon-z` one would execute `wpa_cli select_network 0`.
+The example above was executed through a connection over the wifi interface, hence why the pipe broke.
+
```
CMurtagh-LGTM, line 187:
```suggestion

If the ip mask of the network is not `10.1.255.255` but instead `192.168.255.255` you need to change the static ip of the robot.
Edit `/etc/systemd/networkd/10-bond0.network` and change `10.1.X.X` to `192.168.X.X` keeping the `X` part the same. Now run `# systemctl restart networkd` to apply the changes.
When connecting back to the epsilon network these changes need to be reversed.

```
The ip mask may change, we should describe how to change the static ip.

mrwerdo: Hmm, that's interesting. I was trying to avoid modifying files so that rebooting the system would return it to the default state. I wonder if it's possible to obtain the subnet mask automatically? Maybe through DHCP? Idk.

ysims: This is really only applicable for robocup now, and we will have to modify files anyway to add the robocup networks.
We don't need the spare router anymore, our main router now works for robot and game controller.

mrwerdo:
It sounds like this isn't needed so I'm going to go ahead and close this.

PR #231, opened 2022-07-08
Adds a script to check if any links are broken and prints out the broken link and location in NUbook

David7704:
Checks all links within NUbook point to an existing page, ie no 404 error.


JosephusPaye:
Could you add the script to this PR for review as well?

David7704:
https://raw.githubusercontent.com/David7704/checkLinks/main/checkLinks.js

ysims:
I'm getting this when I try to run it
```
SyntaxError: Cannot use import statement outside a module
```

ysims:
@JosephusPaye I've made some changes here to get it working, it works as intended and should be useful for checking for broken links although not everything it flags is broken. One issue is that the Eigen documentation isn't secure and some DOI links timeout, but they do work. If we add this to CI at some point I'd expect we could just exclude those sites from the checks.

JosephusPaye:
This looks good enough to get the basic job done. A few notes for improvements:

- Make the file operations async and parallel. Right now it's stat-ing and reading one file/directory at a time, blocking everything else while it does that. This should provide a decent speed boost.
- Similarly, making the requests async with a queue that runs a few (say 5) requests in parallel should speed things up too.
- Use the built-in `http` module to make the requests, and avoid adding another dependency.
- Avoid setting max event listeners to `Infinity`. There should be other ways to to solve this if there are warnings about too many event listeners.

> One issue is that the Eigen documentation isn't secure and some DOI links timeout, but they do work. If we add this to CI at some point I'd expect we could just exclude those sites from the checks.

Yeah if this is going to be used in CI a config file can be added with links or domains to ignore. Insecure non-https links should work though right? Or do you get warnings when making requests to those?

ysims:
Seems that all links with error codes outside of 404 seem to be working fine, so at the moment I've changed it to only care about 404 errors. There were issues like the cmake website and the docker website being forbidden :confused:

package.json
JesseWilliamson, line 1: newlines at the end of files

JesseWilliamson (approved):
LGTM

PR #232, opened 2022-07-16
Adds further Git learning resources and docs

BryceTuppurainen:
Unhandled-Runtime-Error

For some reason I'm hitting this error both on main and on this branch at the moment when developing locally, but for the life of me I can't figure out where this would be or whether this is just an issue with the version of nodejs or something that I've got configured at the moment.

Anyway, if you can please check out the content and then merge this in if you can't find anything that you'd suggest fixing that'd be much appreciated.

### DIFF
--- a/src/book/03-guides/04-general/01-learning-resources.mdx
+++ b/src/book/03-guides/04-general/01-learning-resources.mdx
@@ -15,22 +15,25 @@ The majority of resources on this page are free. Some LinkedIn Learning videos a
 ## General
 
 <details>
-<summary>Git</summary>
+<summary> Git </summary>
 
 The NUbots team uses Git and GitHub in its development workflow. Almost any project done on the NUbots team involves using Git and GitHub.
 
 **Reading:**
 
 - [Git and PR Workflow](/guides/general/git) [NUbook]
+- [GitHub Commands Cheat Sheet](https://education.github.com/git-cheat-sheet-education.pdf) [GitHub]
+- [GitHub Official Documentation](https://developer.github.com/) [GitHub]
 
 **Videos:**
 
 - [Git, Git Tools and PR Workflow](https://www.youtube.com/watch?v=sLpEetN3lnk&t=00m00s) [YouTube]
 
-**Tutorials:**
+**Tutorials and Miscellaneous Content:**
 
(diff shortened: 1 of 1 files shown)


BryceTuppurainen:
https://deploy-preview-232--nubook.netlify.app/guides/general/learning-resources <- Deploy Preview 😎

src/book/03-guides/04-general/01-learning-resources.mdx
```diff
@@ -56,7 +59,7 @@ The C++ programming language is used in the main robot codebase. Modern C++ and
 </details>
 
 <details>
-<summary>NUClear</summary>
+<summary> NUClear </summary>
```
LachlanCourt, line 62: Adding a space to these does nothing - if you want to increase the space on either side you'll need to add padding to the summary jsx component. Pls revert these changes 🙏

LachlanCourt (changes requested):
Content looks fine just please fix the formatting

LachlanCourt:
Regarding the runtime error, I am aware of its existence but haven't had time to investigate it. It only occurs in development, and after closing the popup it seems to run fine regardless so other than the extra click it doesn't seem too problematic at this stage. I'll be looking into it more

src/book/03-guides/04-general/01-learning-resources.mdx
BryceTuppurainen: Yes I thought it was strange that there was ignored whitespace in other sections of the document and thought it was for readability initially.

I'll get that corrected and merged in (might also be worth double checking it doesn't appear anywhere else in nubook with some regex to ensure this issue isn't run into again in future)

BryceTuppurainen:
Should be fixed up now, can someone please take a look and then merge this, thanks.

LachlanCourt (approved):
LGTM

JosephusPaye:
> @JosephusPaye do you have any thoughts on this? I think I may have accidentally introduced it during the webpack update

Yeah the update might have introduced the issue. Usually you get the "Unexpected token: <" error when JS tries to import or parse a file that it expects to be JS or JSON, and ends up getting HTML instead. For example, when you import a JS file and the server responds with a 404 HTML page.

PR #234, opened 2022-07-18
Add codedocs link

CMurtagh-LGTM:
Add a link to the codedocs generated by Doxygen.


LachlanCourt (approved):
Looks good to me, might be worth adding a step by step on how to generate, build, and update the docs although that doesn't necessarily need to happen in this PR.

Could go in the system/tools/nubook page or could be a new page system/tools/codedocs if it is quite long

LachlanCourt:
There's a section here that says we will use doxygen in the future we should probably update that at some point too
https://nubook.nubots.net/guides/general/code-conventions#future

LachlanCourt (approved):
LGTM, looks like formatting broke tho lmao

LachlanCourt (approved):
LGTM

PR #237, opened 2022-07-19
Add page on battery handling

LachlanCourt:
[Preview](https://deploy-preview-237--nubook.netlify.app/guides/hardware/batteries)


src/book/03-guides/03-hardware/03-batteries.mdx
```diff
@@ -0,0 +1,62 @@
...
+slug: /guides/hardware/batteries
+authors:
+  - Lachlan Court (@LachlanCourt)
+---
+
+## Battery Safety
+
+<Alert>This page is not designed to be a substitute for the official lab induction, however it should be suitable as a revision tool</Alert>
+
+Our robots run on 4 Cell 14.8V Lithium Polymer (LiPo) batteries. In order to handle batteries in the lab you should have completed a formal lab induction.
```
ysims, line 15: Should mention the Darwin batteries as well.

src/book/03-guides/03-hardware/03-batteries.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+### Charging a battery
+
+1. Plug the battery cable with the white plug into the corresponding socket on the side of the charger
+
+2. Plug the other battery cable into the main output cable of the charger, this may require an adaptor, as seen in the section on using in the robot
+
+3. Plug the Powertech Plus into a power socket, and use the arrow buttons to select the "Charging" setting for a 14.8V 4 Cell battery, as seen in the image below
+
+4. Hold down the start button until the device beeps, and the content on the screen indicates the charging has commenced
```
ysims, line 48: After holding down, check for any errors. If any errors, then don't proceed! If no errors, press start.

src/book/03-guides/03-hardware/03-batteries.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+5. Close and lock the white BATT-SAFE box with the battery inside (Keep the charger outside if possible to hear when it is done)
```
ysims, line 50:
```suggestion
5. Close and lock the white BAT-SAFE box with the battery inside (Keep the charger outside if possible to hear when it is done)
```

src/book/03-guides/03-hardware/03-batteries.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+1. Start by finding the cable on the battery with the white plug. Line up the `-` symbol on the monitor with the 5 black leads of the battery cable, and plug in. Be aware that the battery monitor will beep several times when this happens, this is normal. ![Battery monitor attached to a battery](./images/batteries/battery_monitor.jpg)
+
+2. Once the battery monitor is attached to the battery, the other cable on the battery can now be plugged into the robot. Depending on the battery this may require an adaptor lead, shown below.
+
+3. Once both cables on the battery are hooked up, the robot can safely be turned on. Be sure to carefully load the battery into the robot to avoid damage, this may require some technique so don't be afraid to ask a team member for the best way to go about this.
+
+![Battery in robot with a battery monitor](./images/batteries/battery_in_robot.jpg)
+
+<Alert type="warning">It is important to never use a battery in the robot without a battery monitor, and also to unplug both the robot and the monitor from the battery when not being used. Both of these can lead to the battery going flat beyond recovery</Alert>
```
ysims, line 34:
```suggestion
<Alert type="warning">It is important to never use a battery in the robot without a battery monitor, and also to unplug both the robot and the monitor from the battery when not being used. Both of these can lead to the battery going flat beyond recovery.</Alert>
```

ysims (commented):
Might be good to also mention when we use batteries. I.e. qualification video, any other media video where cables might interfere, lab induction, and RoboCup.

miikyla:
https://www.jaycar.com.au/medias/sys_master/images/images/9289572286494/MB3632-manualMain.pdf

src/book/03-guides/03-hardware/03-batteries.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+## Charging
+
+We have two types of chargers in the lab to charge LiPo batteries. Both achieve the same purpose and have a similar user interface. The Powertech Plus charger will be used for the purposes of the images below.
+
+### Charging a battery
+
+1. Plug the battery cable with the white plug into the corresponding socket on the side of the charger
+
```
miikyla, line 43: I have a feeling there's a couple more steps you need to go through from scratch (which are usually already selected when you chuck on one of our batteries). Would be valuable to include these (from memory it's like chemistry selection stuff), just in case something has been accidentally misconfigured

miikyla (changes requested):

CMurtagh-LGTM:
We should also have information on disposal, I think the information is on the induction sheet.

ysims:
> We should also have information on disposal, I think the information is on the induction sheet.

Good idea. EE handle battery disposal for us - the bucket for batteries is in room EEG07, or take them to Ian Powell in EEG08.

src/book/03-guides/03-hardware/03-batteries.mdx
Bidski: Split this up into two alerts so that you are less likely to gloss over important information while reading
```suggestion
<Alert type="warning">Never use a battery in the robot without a battery monitor</Alert>

<Alert type="warning">Always unplug both the robot and the monitor from the battery when not being used. Both of these can lead to the battery going flat beyond recovery</Alert>
```

src/book/03-guides/03-hardware/03-batteries.mdx
```diff
@@ -0,0 +1,62 @@
...
+5. Close and lock the white BATT-SAFE box with the battery inside (Keep the charger outside if possible to hear when it is done)
+
+6. Stay in the lab with the battery and disconnect immediately when charge has completed. Never leave a battery charging when there is no one in the lab
+
+![Charging setting on Powertech Plus LiPo Charger](./images/batteries/charge.jpg)
+
+## Storing and Traveling with batteries
+
+When travelling, LiPo batteries should always go in carry-on luggage in aircraft, never in checked luggage. It is important to prepare batteries for storage when travel in aircraft. Batteries should also be prepared for storage when being stored in the lab for an extended period of time.
```
Bidski, line 58:
```suggestion
When travelling, LiPo batteries should always be kept in fire retardant bags, with one battery stored in each fire retardant bag. If travelling on an aircraft LiPo batteries 
should always go in carry-on luggage, never in checked luggage. It is important to prepare batteries for storage when travelling in an aircraft. Batteries should also be prepared for storage when being stored in the lab for an extended period of time. 
```

Bidski (changes requested):

src/book/03-guides/03-hardware/03-batteries.mdx
LachlanCourt: I reckon if that's the case then they can look up the documentation themselves. This shows what it should look like in a standard case and it's not designed to be a setup/troubleshooting guide so unless we cover every case I feel like this should be suitable imo :thinking:

src/book/03-guides/03-hardware/03-batteries.mdx
```diff
@@ -0,0 +1,86 @@
...
+authors:
+  - Lachlan Court (@LachlanCourt)
+---
+
+## Battery Safety
+
+<Alert>This page is not designed to be a substitute for the official lab induction, however it should be suitable as a revision tool</Alert>
+
+Our robots have the capability to run on both battery power and by plugging into a power supply. The robots should be run on battery power for events such as
+RoboCup, RoboCup qualification videos, lab inductions, and other times were a cable attached to the robot may intefere with its operation. During standard testing in
```
Bidski, line 16:
```suggestion
RoboCup, RoboCup qualification videos, lab inductions, and other times where a cable attached to the robot may interfere with its operation. During standard testing in
```

src/book/03-guides/03-hardware/03-batteries.mdx
```diff
@@ -0,0 +1,86 @@
...
+
+## Using in the robot
+
+<Alert type="warning">Never use a battery in the robot without a battery monitor</Alert>
+
+<Alert type="warning">Always unplug both the robot and the monitor from the battery when not being used. Both of these can lead to the battery going flat beyond
+recovery</Alert>
+
+Whenever a battery is being used in the robot, it should be additionally connected to a battery monitor. This ensures that the battery will not run so low that it
+will be unusably flat. Unlike regular batteries, LiPo batteries cannot run completely flat unless they are due to be disposed of.
```
Bidski, line 34: The monitor "ensures" nothing. It only provides an audible warning when the battery is flat.

src/book/03-guides/03-hardware/03-batteries.mdx
```diff
@@ -0,0 +1,86 @@
...
+To connect a battery monitor to the battery:
+
+1. Start by finding the cable on the battery with the white plug. Line up the `-` symbol on the monitor with the 5 black leads of the battery cable, and plug in. Be
+   aware that the battery monitor will beep several times when this happens, this is normal. ![Battery monitor attached to a battery](./images/batteries/battery_monitor.jpg)
+
+2. Once the battery monitor is attached to the battery, the other cable on the battery can now be plugged into the robot. Depending on the battery this may require an
+   adaptor lead, shown below.
+
+3. Once both cables on the battery are hooked up, the robot can safely be turned on. Be sure to carefully load the battery into the robot to avoid damage, this may
+   require some technique so don't be afraid to ask a team member for the best way to go about this.
```
Bidski, line 45: Split these up into 2 points
```suggestion
3. Once both cables on the battery cables are hooked up, install the battery into the robot and install the back panel cover. Be sure to carefully load the battery into the robot to avoid damage, this may require some technique so don't be afraid to ask a team member for the best way to go about this.
4. The robot can now be safely turned on.
```

src/book/03-guides/03-hardware/03-batteries.mdx
```diff
@@ -0,0 +1,86 @@
...
+![Battery in robot with a battery monitor](./images/batteries/battery_in_robot.jpg)
+
+## Charging
+
+We have two types of chargers in the lab to charge LiPo batteries. Both achieve the same purpose and have a similar user interface. The Powertech Plus charger will be
+used for the purposes of the images below.
+
+### Charging a battery
+
+1. Plug the battery cable with the white plug into the corresponding socket on the side of the charger
```
Bidski, line 56: Might be worth pointing out that these white sockets are keyed and the plug on the battery should only fit into one of the sockets and if you have to force it you are doing it wrong

src/book/03-guides/03-hardware/03-batteries.mdx
Bidski: Then mention that if the interface differs to what is shown in the pictures that the help of a team member (and/or consultation with the setup/troubleshooting guide) should be sought as blindly going ahead could result in catastrophic battery failure (the chargers are usually good enough to detect bad cases, so "catastrophic" is probably over-the-top, but you get the point)

src/book/03-guides/03-hardware/03-batteries.mdx
```diff
@@ -0,0 +1,86 @@
...
+
+2. Plug the other battery cable into the main output cable of the charger, this may require an adaptor, as seen in the section on using in the robot
+
+3. Plug the Powertech Plus into a power socket, and use the arrow buttons to select the "Charging" setting for the type of battery you a charging. For the NUgus robot
```
Bidski, line 60:
```suggestion
3. Plug the Powertech Plus into a power socket, and use the arrow buttons to select the "Charging" setting for the type of battery you are charging. For the NUgus robot
```

src/book/03-guides/03-hardware/03-batteries.mdx
```diff
@@ -0,0 +1,86 @@
...
+
+6. Stay in the lab with the battery and disconnect immediately when charge has completed. Never leave a battery charging when there is no one in the lab
+
+![Charging setting on Powertech Plus LiPo Charger](./images/batteries/charge.jpg)
+
+## Storing and Traveling with batteries
+
+When travelling, LiPo batteries should always be kept in fire retardant bags, with one battery stored in each fire retardant bag. If travelling on an aircraft,
+LiPo batteries should always go in carry-on luggage, never in checked luggage. It is important to prepare batteries for storage when travel in aircraft. Batteries
+should also be prepared for storage when being stored in the lab for an extended period of time.
```
Bidski, line 76: Care should be taken to verify LiPo battery limits that your airlines have set. They will usually specify a maximum `Ah` value for all batteries and possibly a maximum number of batteries. If they catch you carrying more than this limit they will probably make you throw the batteries out

src/book/03-guides/03-hardware/03-batteries.mdx
```diff
@@ -0,0 +1,86 @@
...
+
+In both cases, preparing a battery for storage is done by following the same process outlined for charging, however by using the "Storage" setting, as seen in the
+image below.
+
+![Storage setting on Powertech Plus LiPo Charger](./images/batteries/storage.jpg)
+
+## Disposal of batteries
+
+If a battery has been physically damaged or is showing errors when attempting to charge, it should be disposed of. Batteries for disposal should be taken to the
+Electrical Engineering building on campus. Talk to the team leader to confirm the most up to date details of where to take them.
```
Bidski, line 86: Acquiring some electrical tape so you can "bandage" damaged cables (so they dont short out when in the fire retardant bag might be a good idea too)

Bidski (changes requested):

Bidski (dismissed):
LGTM

src/book/03-guides/03-hardware/03-batteries.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+6. Stay in the lab with the battery and disconnect immediately when charge has completed. Never leave a battery charging when there is no one in the lab
+
+![Charging setting on Powertech Plus LiPo Charger](./images/batteries/charge.jpg)
+
+## Storing and Traveling with batteries
+
+<Alert>
+
+Care should be taken to verify the LiPo battery limits of the airline you are travelling with. They will usually specify a maximum `Ah` value for all batteries and
```
miikyla, line 82: They calculate in `Wh` not `Ah`, which is calculated using `Voltage * Ah`

src/book/03-guides/03-hardware/03-batteries.mdx
```diff
@@ -0,0 +1,101 @@
...
+image below.
+
+![Storage setting on Powertech Plus LiPo Charger](./images/batteries/storage.jpg)
+
+## Disposal of batteries
+
+If a battery has been physically damaged or is showing errors when attempting to charge, it should be disposed of. Batteries for disposal should be taken in a fire
+retardant bag to the Electrical Engineering building on campus. Talk to the team leader to confirm the most up to date details of where to take them.
+
+If the cables on the battery are damaged, it is a good idea to wrap the damaged section in electrical tape to prevent them short cicruiting inside the fire retardant
```
miikyla, line 100: Specify that the single cable must be wrapped, it could be interpreted that the entire section needs to be wrapped together with this wording (which would obvs cause the short then)

miikyla (changes requested):
Couple of things, otherwise it's g

src/book/03-guides/03-hardware/03-batteries.mdx
miikyla: https://www.qantas.com/au/en/travel-info/baggage/dangerous-goods/spare-batteries-and-powerbanks.html#:~:text=Note%3A%20Watt%20hours%20(Wh)%20are%20determined%20by%20multiplying%20the%20voltage%20(V)%20by%20the%20amp%20hours%20(Ah).%20ie.%2012V%20x%205Ah%20%3D%2060Wh

miikyla: I distinctly remember being very confused when they asked for Watt Hours since it's such a weird unit lol

miikyla (approved):
:shipit:

PR #238, opened 2022-07-20
Make SVG icons renderable in github

Bidski:
Came across this today and it bugs in a way that I can't explain

No idea why they still don't display in the PR view, but viewing the files themselves (click the `...` and then `View File` will allow the files to be rendered properly)


LachlanCourt:
There are like 30 svgs in NUbook was there supposed to be more files here?

Bidski:
> There are like 30 svgs in NUbook was there supposed to be more files here?

I had a look at the others, these are the only 3 that I saw that didnt render in github/browser.

JosephusPaye:
Yeah, the rest have the `xmlns` attribute, which appears to be necessary and sufficient to render them in the GitHub web UI.

@Bidski I think for consistency with the rest of those icons we should only add that attribute to these 3, and exclude the `<?xml` line.

PR #248, opened 2022-08-09
Update yarn Ubuntu installation instructions

thomaslegge98:
Update yarn installation instructions to match what is used on NUsight page for Ubuntu https://nubook.nubots.net/guides/main/getting-started#get-build-and-run-nusight
There are issues with installing yarn using current version.


src/book/02-system/03-tools/02-nubook.mdx
```diff
@@ -28,7 +28,13 @@ You'll need [Git](https://git-scm.com/), [Node.js](https://nodejs.org/en/) and [
 
 2. Install the latest version of Node.js by following the instructions [here](https://github.com/nodesource/distributions/blob/master/README.md#installation-instructions) (pick the latest version).
```
ysims, line 29:
```suggestion
2. Install Node.js version 14 by following the instructions [here](https://github.com/nodesource/distributions/blob/master/README.md#installation-instructions) and running the commands under "Node.js v14.x".
```

since the latest doesn't seem to work

thomaslegge98: Yes. Was thinking to put that but got slack, absolutely should.

ysims:
The checks failed, if you click on the details link it'll take you to the logs which are complaining about formatting. Run `yarn format` and commit the formatting changes and it should be good.

src/book/02-system/03-tools/02-nubook.mdx
```diff
@@ -26,9 +26,15 @@ You'll need [Git](https://git-scm.com/), [Node.js](https://nodejs.org/en/) and [
    sudo apt install git
    ```
 
-2. Install the latest version of Node.js by following the instructions [here](https://github.com/nodesource/distributions/blob/master/README.md#installation-instructions) (pick the latest version).
+2. Install Node.js version 14 by following the instructions [here](https://github.com/nodesource/distributions/blob/master/README.md#installation-instructions) and running the commands under "Node.js v14.x".
 
-3. Install the latest version of Yarn by following the instructions [here](https://classic.yarnpkg.com/en/docs/install#debian-stable).
+3. Install the latest version of Yarn
```
LachlanCourt, line 31: You should just be able to do `npm -g install yarn` after installing Node 🤔 is there a reason you are installing yarn directly from source? Installing a specific version of Node and then the latest version of yarn could potentially introduce a conflict whereas npm will manage the version properly

src/book/02-system/03-tools/02-nubook.mdx
```diff
@@ -26,9 +26,15 @@ You'll need [Git](https://git-scm.com/), [Node.js](https://nodejs.org/en/) and [
    sudo apt install git
    ```
 
-2. Install the latest version of Node.js by following the instructions [here](https://github.com/nodesource/distributions/blob/master/README.md#installation-instructions) (pick the latest version).
+2. Install Node.js version 14 by following the instructions [here](https://github.com/nodesource/distributions/blob/master/README.md#installation-instructions) and running the commands under "Node.js v14.x".
```
LachlanCourt, line 29: I believe this should work with Node 16, will double check in the lab today. 16 is the current LTS so we should stick to that if possible

src/book/02-system/03-tools/02-nubook.mdx
LachlanCourt: API has slightly changed but yeah the way I would do this would be
`curl -fsSL https://deb.nodesource.com/setup_16.x | sudo -E bash -
sudo apt-get install -y nodejs
npm install yarn --location=global`

LachlanCourt: Might be worth updating the instructions here as well https://github.com/NUbots/NUbook/blob/main/src/book/03-guides/01-main-codebase/01-getting-started.mdx

src/book/02-system/03-tools/02-nubook.mdx
LachlanCourt: Node 16 works too, I reckon we go with that

src/book/02-system/03-tools/02-nubook.mdx
thomaslegge98: `npm -g install yarn` wasn't working for me for some reason and was able to with `npm install --global yarn
` I will go through the setup again tonight and verify which commands work / don't.

src/book/02-system/03-tools/02-nubook.mdx
thomaslegge98: Okay, I will go through the setup again and see if I can get the later versions to work.

src/book/02-system/03-tools/02-nubook.mdx
LachlanCourt: The npm one may need to run as sudo

thomaslegge98: Okay, so after I went from Pop_OS version 22.0x -> 22.0x, the original installation works.

src/book/02-system/03-tools/02-nubook.mdx
thomaslegge98: You're correct, Node 16 worked for me as well.

src/book/02-system/03-tools/02-nubook.mdx
thomaslegge98: Whatever is currently there written works, provided you install Node 16 over 18. So I will update it.

ysims (approved):
I think this is fine, it does the job (stops people getting the wrong version). If we want to change the instances where we link to a page to instead putting the lines to run on NUbook, can be done in a follow up PR with changing the getting started page too. It would only be for cases where we are extremely certain the steps won't change.

PR #250, opened 2022-08-13
Remove image from other folder that shouldn't exist anymore

ysims:
The image still exists in `/system/tools/images`.


JosephusPaye (approved):
LGTM

Issue #224, opened 2022-06-06
Update Algolia to v3


LachlanCourt:
New version of search

LachlanCourt:
See Joey's comment here
https://github.com/NUbots/NUbook/pull/221#issuecomment-1143406162

PR #252, opened 2022-08-14
Fix unexpected token error 

LachlanCourt:
Fixes unexpected token and duplicate script rendering. This should all be fixed when I finish the upgrade to algolia 3 but until I finish tidying up that this will at least stop the runtime errors when running NUbook locally, will hopefully get the other one up in the next week or two

[Preview](https://deploy-preview-252--nubook.netlify.app/)


ysims (approved):
sure! maybe if there's some further things to do make an issue unless that's covered by the search upgrade stuff

LachlanCourt:
Should all be made obsolete after the [proper upgrade](https://github.com/NUbots/NUbook/issues/224) this is just a patch for now

PR #261, opened 2022-09-20
Update the maths symbol convention to include unit vectors and acceleration

TrentHouliston:


src/book/02-system/01-foundations/05-mathematics.mdx
```diff
@@ -201,18 +201,50 @@ This is written in the codebase as
...
 ```
 
-### Velocity
+### Unit Vectors
 
-The velocity of some object is represented relative to some space. If we have the **velocity** of object A moving **relative** to space b, then we write this in the codebase as `vAb`.
+A **unit vector** is a vector with a **length** of **1**. They are used to represent **directions** in space.
 
-Velocities can only be **rotated**, **not translated**. They should **not** be multiplied by a homogeneous transformation matrix, but they can be multiplied by a **rotation matrix**.
+They are written by describing their origin the direction they are pointing and the space they are in.
```
LachlanCourt, line 208:
```suggestion
They are written by describing their origin, the direction they are pointing and the space they are in.
```

LachlanCourt (approved):
This all seems fine to me, just missing that comma, pls run `yarn format` to appease the build tests too 🙏

ysims (approved):

PR #265, opened 2022-10-19
Add instructions for changing robocup network.

CMurtagh-LGTM:
https://github.com/NUbots/NUbots/pull/911


src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
```diff
@@ -0,0 +1,27 @@
...
+
+1. `cd system/nugusX/etc/systemd/network/` to go to the config folder that system configuration will use for networkd.
+2. Rename `40-wlp58s0-robocup.network` to `30-wlp58s0-robocup.network`
+3. Rename `30-wlp58s0.network` to `40-wlp58s0.network`
+4. Double check that network, gateway and dns are correct.
+   Usually the address is `192.168.0.X` or `10.1.0.X` and both gateway and dns end with `3.1`.
+5. `cd ~` to go back to home.
+5. Add the output of `wpa_passphrase SSID PASSWORD` to `system/default/etc/wpa_supplicant/wpa_supplicant-wlp58s0.conf`
+   and add `priority=3` inside the curly brackets on a new line.
+   You can use 😎 `vim` or 🤮 `nano` to edit files on the robot.
```
ysims, line 24:
```suggestion
   You can use `vim` or `nano` to edit files on the robot.
```

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
```diff
@@ -0,0 +1,27 @@
...
+
+## Networking
+
+1. `cd system/nugusX/etc/systemd/network/` to go to the config folder that system configuration will use for networkd.
+2. Rename `40-wlp58s0-robocup.network` to `30-wlp58s0-robocup.network`
+3. Rename `30-wlp58s0.network` to `40-wlp58s0.network`
+4. Double check that network, gateway and dns are correct.
+   Usually the address is `192.168.0.X` or `10.1.0.X` and both gateway and dns end with `3.1`.
+5. `cd ~` to go back to home.
+5. Add the output of `wpa_passphrase SSID PASSWORD` to `system/default/etc/wpa_supplicant/wpa_supplicant-wlp58s0.conf`
```
ysims, line 22: This wouldn't make sense to a lot of people, particularly those running it for the first time. Might as well just say to run 
`wpa_passphrase SSID PASSWORD > system/default/etc/wpa_supplicant/wpa_supplicant-wlp58s0.conf` 
you probably should do an absolute path as well. and say what SSID and PASSWORD meanm and what the whole line does.

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
```diff
@@ -0,0 +1,27 @@
...
+   and add `priority=3` inside the curly brackets on a new line.
```
ysims, line 23: probably specify to look for the new entry with the SSID input just before

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
```diff
@@ -0,0 +1,27 @@
...
+section: Guides
+chapter: Main Codebase
+title: Robocup Setup
+description: Setting up the robot for robocup
+slug: /guides/main/getting-started
+authors:
+  - Ysobel Sims (@ysims)
+  - Josephus Paye II (@JosephusPaye)
+  - Mikyla Peters (@miikyla)
+  - Lachlan Court (@LachlanCourt)
```
ysims, line 11:
```suggestion
  - Cameron Murtagh (@CMurtagh-LGTM)
```

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
```diff
@@ -0,0 +1,27 @@
...
+---
+
+## Networking
+
+1. `cd system/nugusX/etc/systemd/network/` to go to the config folder that system configuration will use for networkd.
+2. Rename `40-wlp58s0-robocup.network` to `30-wlp58s0-robocup.network`
+3. Rename `30-wlp58s0.network` to `40-wlp58s0.network`
```
TrentHouliston, line 18: You could just rename the RoboCup one to be lower than 30 rather than renaming both. Also I think both the old 30 file and the new 30 file will both be around still, and it'll just be lexicographical order of which one runs.

Also it might be worth adding something as to how to undo this after RoboCup, since both of these files will be left behind and you'll need to put it back in "Lab" mode

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
```diff
@@ -0,0 +1,27 @@
+---
+section: Guides
+chapter: Main Codebase
+title: Robocup Setup
+description: Setting up the robot for robocup
+slug: /guides/main/getting-started
```
ysims, line 6:
```suggestion
slug: /guides/main/robocup-network
```

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
```diff
@@ -0,0 +1,27 @@
+---
+section: Guides
+chapter: Main Codebase
+title: Robocup Setup
```
ysims, line 4:
```suggestion
title: RoboCup Network Setup
```

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
```diff
@@ -0,0 +1,27 @@
+---
+section: Guides
+chapter: Main Codebase
```
ysims, line 3: It's not quite this. More firmware. Maybe under hardware, since it's most close to Flashing a Robot? Not sure. It could even go in that page.

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
```diff
@@ -0,0 +1,27 @@
...
+title: Robocup Setup
+description: Setting up the robot for robocup
```
ysims, line 5:
```suggestion
description: Setting up the robot networking for robocup
```

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
TrentHouliston: boooo :P

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
TrentHouliston: I think the best way might be to swap the two files rather than changing their numbers actually...

ysims:
Probably want the files changed on the branch so they aren't overwritten with the usual config when the robots are built to at robocup

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
TrentHouliston: If wpa_supplicant is in the docker image you could just do something with ./b shell

CMurtagh-LGTM:
Maybe say something about the subnet.

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
CMurtagh-LGTM: I was thinking we could have more than just this guide on this page :shrug:.

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
CMurtagh-LGTM: Swapping them will be good, the system configuration tool does not delete any old files in these directories so they will stay around on the robot.

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
```diff
@@ -0,0 +1,27 @@
...
+  - Lachlan Court (@LachlanCourt)
+---
+
+## Networking
+
+1. `cd system/nugusX/etc/systemd/network/` to go to the config folder that system configuration will use for networkd.
+2. Rename `40-wlp58s0-robocup.network` to `30-wlp58s0-robocup.network`
+3. Rename `30-wlp58s0.network` to `40-wlp58s0.network`
+4. Double check that network, gateway and dns are correct.
+   Usually the address is `192.168.0.X` or `10.1.0.X` and both gateway and dns end with `3.1`.
```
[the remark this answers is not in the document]
CMurtagh-LGTM: `10.1.X.X` and `192.168.X.X` are kinda universal, basically all default subnets are one or the other.

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
CMurtagh-LGTM: I don't want people blindly appending to the `wpa_supplicant` config file, I want to make sure they look at it before they add stuff.

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
```diff
@@ -0,0 +1,34 @@
...
+title: Robocup Setup
+description: Setting up the robot for robocup
+slug: /guides/main/robocup-setup
+authors:
+    - Cameron Murtagh (@CMurtagh-LGTM)
+---
+
+## Networking
+
+1. On the robot with a monitor and keyboard, not via `ssh`.
```
LachlanCourt, line 13: Maybe have this as an Alert as it's not really a step but more a method of operation
```suggestion
<Alert type='info'>

Network setup must be completed on the robot with a monitor and keyboard, not via `ssh`.

</Alert>
```

TrentHouliston: If you do this right you can do it by ssh. We always used to. Only had to go to mouse and keyboard when we screwed up :P

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
```diff
@@ -0,0 +1,34 @@
...
+2. Rename `system/nugusX/etc/systemd/network/40-wlp58s0-robocup.network`
+   to `system/nugusX/etc/systemd/network/30-wlp58s0.network`
+3. Rename `system/defualt/etc/systemd/30-wlp58s0.network` to `system/default/etc/systemd/40-wlp58s0-robocup.network`
```
TrentHouliston, line 16:
```suggestion
3. Rename `system/default/etc/systemd/30-wlp58s0.network` to `system/default/etc/systemd/40-wlp58s0-robocup.network`
```
You don't actually have to do this step though. You can just rename the robocup one to the 30 one

Also should really do this in the codebase on a RoboCup branch so it persists and gets set every time.
I know you mention this later but I think that should be the normal way of dealing with robocup.
Have the config setup on a RoboCup branch for each robot and just run from there

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
```diff
@@ -0,0 +1,34 @@
...
+4. Double check that `Address`, in the file now called `30-wlp58s0.network`, is in the subnet you are connecting to,
+   this is the address of this robot.
+   Usually the subnet mask is `255.255.0.0` meaning that the first two numbers define the subnet.
+   Usually the address of the subnet is `192.168.X.X` or `10.1.X.X`.
+5. Also check that `Gateway` and the first `DNS` are correct in this file, both gateway and dns normally end with
+   `.3.1` and are in the subnet.
```
TrentHouliston, line 22: Normally the RoboCup networks don't have a DNS or gateway

Bidski, line 22: Isn't `.3.1` highly specific to the network you are connecting to? I feel like saying "normally" here is misleading.

DNS is also irrelevant for the robocup network (there is no internet to connect to and all communication on this network is either broadcast/multicast or to a specific IP).

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
```diff
@@ -0,0 +1,34 @@
...
+6. Add the output of `wpa_passphrase SSID PASSWORD` to `system/default/etc/wpa_supplicant/wpa_supplicant-wlp58s0.conf`
```
Bidski, line 23:
```sh
wpa_passphrase SSID PASSWORD >> system/default/etc/wpa_supplicant/wpa_supplicant-wlp58s0.conf
```

Or just manually add the entire block to the file
```
network={
       ssid="SSID"
       psk="PASSWORD"
       priority=3
}
```

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
```diff
@@ -0,0 +1,34 @@
...
+   where `SSID` is the SSID and `PASSWORD` is the password of the Wi-Fi network you are connecting to.
+   Now add `priority=3` inside the curly brackets on a new line to the entry that you just added.
```
Bidski, line 25: I feel like you should show an example config file here so people know what it should look like

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
```diff
@@ -0,0 +1,34 @@
...
+   You can use `vim` or `nano` to edit files on the robot.
+7. Run `sudo ./system_configuration`.
+8. Reboot the robot `reboot` or restart the services
+   `sudo systemctl restart wpa_supplicant@wlp58s0` and `sudo systemctl restart wpa_supplicant@wlp58s0`.
```
Bidski, line 29: If the network file has also been changed (`/etc/systemd/40-wlp58s0-robocup.network`) you probably also need to do a `sudo systemctl daemon-reload` followed by a `sudo systemctl restart systemd-networkd` before you restart `wpa_supplicant`. Might be a good idea to make `system_configuration` restart the services itself?

Bidski (changes requested):

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
CMurtagh-LGTM: I think it is clearer if we do *something* with the default one.

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
```diff
@@ -10,25 +10,35 @@ authors:
...
+           priority=3
+    }
+    ```
+4.  For each robot `X` do steps 5 to 10.
+5.  Rename `nugusX/etc/systemd/network/40-wlp58s0-robocup.network`
+    to `nugusX/etc/systemd/network/30-wlp58s0.network`.
+6.  Double check that `Address`, in the file now called `30-wlp58s0.network`, is in the subnet you are connecting to,
+    this is the address of this robot.
+    Usually the subnet mask is `255.255.0.0` meaning that the first two numbers define the subnet.
+    Usually the address of the subnet is `192.168.x.x` or `10.1.x.x`.
```
TrentHouliston, line 33: Might be worth saying that the specific IP addresses and subnet masks are provided by RoboCup.
Often we end up with a `192.168.TEAM_NUMBER.ROBOTNUMBER` but different years the format has varied

TrentHouliston (dismissed):

CMurtagh-LGTM:
C.I. fails due to #266

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
```diff
@@ -5,30 +5,40 @@ title: Robocup Setup
...
-   where `SSID` is the SSID and `PASSWORD` is the password of the Wi-Fi network you are connecting to.
-   Now add `priority=3` inside the curly brackets on a new line to the entry that you just added.
-   You can use `vim` or `nano` to edit files on the robot.
-7. Run `sudo ./system_configuration`.
-8. Reboot the robot `reboot` or restart the services
-   `sudo systemctl restart wpa_supplicant@wlp58s0` and `sudo systemctl restart wpa_supplicant@wlp58s0`.
+1.  Checkout the branch for robocup and change directory to `modules/tools/SystemConfiguration/data/system`.
+2.  Rename `default/etc/systemd/30-wlp58s0.network` to `default/etc/systemd/40-wlp58s0-robocup.network`
+3.  Add the output of `wpa_passphrase SSID PASSWORD` to `default/etc/wpa_supplicant/wpa_supplicant-wlp58s0.conf`
+    where `SSID` is the SSID and `PASSWORD` is the password of the Wi-Fi network you are connecting to.
```
Bidski, line 16:
```suggestion
    where `SSID` and `PASSWORD` are the SSID and the password of the Wi-Fi network you are connecting to.
```

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
```diff
@@ -5,30 +5,40 @@ title: Robocup Setup
...
+    Now add `priority=3` inside the curly brackets on a new line to the entry that you just added.
+    The block should look like, where `HASHED_PASSWORD` is a long string of letters and numbers
+    ```
+    network={
+           ssid="SSID"
+           #psk="PASSWORD"
+           psk=HASHED_PASSWORD
+           priority=3
+    }
+    ```
```
Bidski, line 26: I still thing you should just tell people to add this block to the end of `default/etc/wpa_supplicant/wpa_supplicant-wlp58s0.conf`, they have to come open this file in an editor to add the `priority=3`line anyway, and telling them to add the whole block to the end of the file circumvents people accidentally adding the priority line to the wrong block 
```suggestion
    ```
    network={
           ssid="SSID"
           psk="PASSWORD"
           priority=3
    }
    ```
```

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
```diff
@@ -5,30 +5,40 @@ title: Robocup Setup
...
-and when rerunning steps `7` and `8` it will set the network to what the files in steps `1-6` indicate.
-You can create a robocup branch, on your computer, with these files edited in the
-`modules/tools/SystemConfiguration/data/` folder, so it will not be overwritten when installing.
+    When running via `ssh` make sure that you have a monitor and keyboard handy to handle any mistakes.
+
+    </Alert>
+
+9.  Run `sudo ./system_configuration` on the robot.
+10. Reboot the robot `reboot` or restart the services `sudo systemctl daemon-reload`,
+    `sudo systemctl restart systemd-networkd` and `sudo systemctl restart wpa_supplicant@wlp58s0`.
```
Bidski, line 44:
```suggestion
10. Reboot the robot `reboot` or restart the services `sudo systemctl daemon-reload && sudo systemctl restart systemd-networkd && sudo systemctl restart wpa_supplicant@wlp58s0`.
```

Bidski (changes requested):

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
CMurtagh-LGTM: I'm not a fan of just slapping all three commands together, commands linked together with lots of `&&` are quite ugly and, I feel, make it look more complex.

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
ysims: Does this work? I remember trying to manually put in the info once before and it wouldn't work until I added it in with `wpa_passphrase` and had the hashed password? But maybe I did something else wrong :shrug:

Bidski: The manpages wpa_supplicant.conf specifically say
```
psk          WPA preshared key used in WPA-PSK mode.  The key is specified as
	     64	hex digits or as an 8-63 character ASCII passphrase.  ASCII
	     passphrases are dynamically converted to a	256-bit	key at runtime
	     using the network SSID, or	they can be statically converted at
	     configuration time	using the wpa_passphrase(8) utility.
```
And this forum post also says this
https://superuser.com/a/679959/1676555

so it should work just fine

ysims:
I did have one issue when using this guide the other day. The bond network was up, and so it wasn't working because bond settings are the lab settings. I added something in about updating the bond file as well but it seems like it'd be easier to just update bond and not touch the wlp files. Would this cause problems if bond isn't up? I imagine not because the default settings have wlp network=bond0. It seems like it'd be common for bond to be up (I don't know how it works but I suspect if ethernet and wireless are used it comes on?) during robocup.

src/book/03-guides/01-main-codebase/04-robocup-setup.mdx
ysims: this would be better, I'll test and update it

CMurtagh-LGTM:
> 

The idea was that you can still connect via ethernet with the same ip when the wifi is on the robocup settings.

ysims:
I've added the robocup configuration tool into this PR, and addressed reviews for the manual component (as well as going through it myself).

src/book/03-guides/04-general/03-robocup-setup.mdx
```diff
@@ -0,0 +1,179 @@
...
+| Network       | <kbd>N</kbd>     | Writes to the configuration files and restarts the network services.                                                                                                                                                                                                                     |
+| Start RoboCup | <kbd>S</kbd>     | Starts the RoboCup service, which runs the `robocup` binary, including on restart.                                                                                                                                                                                                       |
+| Stop RoboCup  | <kbd>D</kbd>     | Stops the RoboCup service, and should be done when a robot is finished with a game. If this is not done then the next time someone uses the robot it will move. This can also cause robots out of the game to interfere with other robots behaviour due to robot-to-robot communication. |
```
jpptm, line 67: This sounds like something we could automate. I think systemd should have a thing for this so that when we ctrl c and exit the binary manually, the robocup service is turned off as well. From a quick search it seems that KillMode=control-group with ExecStopPost=systemctl disable robocup.service is the thing for this? @Bidski is what Im yapping about in the right track though? This comment is just a note for me or anyone who wants to implement this change

src/book/03-guides/04-general/03-robocup-setup.mdx
```diff
@@ -0,0 +1,179 @@
...
+       </Alert>
+
+    6. Run `sudo ./systemconfiguration` on the robot.
+
+    7. Restart the network services on the robot:
+
+       ```bash
+       sudo systemctl daemon-reload
+       sudo systemctl restart systemd-networkd
+       sudo systemctl restart wpa_supplicant@wlp58s0
```
jpptm, line 150:
```suggestion
       sudo systemctl restart systemd-networkd wpa_supplicant@wlp58s0
```

src/book/03-guides/04-general/03-robocup-setup.mdx
```diff
@@ -0,0 +1,179 @@
...
+2. Run `./b build`.
+
+3. For each robot, run `./b install -co <ADDRESS>`.
+
+4. On each robot, run `sudo ./systemconfiguration` and restart the network services.
+
+   ```bash
+   sudo systemctl daemon-reload
+   sudo systemctl restart systemd-networkd
+   sudo systemctl restart wpa_supplicant@wlp58s0
```
jpptm, line 166:
```suggestion
   sudo systemctl restart systemd-networkd wpa_supplicant@wlp58s0
```

jpptm (changes requested):

src/book/03-guides/04-general/03-robocup-setup.mdx
```diff
@@ -0,0 +1,179 @@
...
+
+The RoboCup configuration tool involves an ncurses interface for easy set up of the robot in a game. It contains sections for network settings and general game configuration.
+
+To run the RoboCup configuration tool, run `./b configure -i` and turn on `ROLE_robocupconfiguration`. Run `./b build` to build the binary.
+
+It is possible to run `./b run robocupconfiguration` locally to see the tool, but it is best used on a robot.
+
+Install the files to robot number `X` with `./b install -co nX`. On the robot, run `sudo ./robocupconfiguration`. `sudo` is required if changing any network settings.
+
+The following code snippit shows what the ncurses terminal interface used in the RoboCup configuration tool looks like.
```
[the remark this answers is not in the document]
ysims: I think it's better to have a code block than an image, as its more accessible. 
We do the same thing here https://nubook.nubots.net/guides/main/tuning-and-running-scripts

src/book/03-guides/04-general/03-robocup-setup.mdx
```diff
@@ -0,0 +1,179 @@
...
+
+## RoboCup Configuration Tool
+
+The RoboCup configuration tool involves an ncurses interface for easy set up of the robot in a game. It contains sections for network settings and general game configuration.
+
+To run the RoboCup configuration tool, run `./b configure -i` and turn on `ROLE_robocupconfiguration`. Run `./b build` to build the binary.
+
+It is possible to run `./b run robocupconfiguration` locally to see the tool, but it is best used on a robot.
+
+Install the files to robot number `X` with `./b install -co nX`. On the robot, run `sudo ./robocupconfiguration`. `sudo` is required if changing any network settings.
```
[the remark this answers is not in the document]
ysims:
```suggestion
Install the files to robot number `X` with `./b install -co nX`. Alternatively, `./b install -co -u nubots <robot's IP address>` if `X` is unknown.

On the robot, run `sudo ./robocupconfiguration`. `sudo` is required if changing any network settings.
```

src/book/03-guides/04-general/03-robocup-setup.mdx
ysims: There was also the idea (I think from Dex) about having a USB that controlled the service being on/off.

jpptm: IMO systemd solution leaves us an extra port to use which is slightly as it doesnt rely on a phys device

jpptm (approved):
lgtm

src/book/03-guides/04-general/03-robocup-setup.mdx
ysims: I'm not sure how it would help though, we don't have any ssh connection to the robot during a game to stop the binary manually. It wouldn't solve the issue of the robot moving when it turns on, as the problem is that people forget to stop it before turning the robot off after a game or when doing repairs. With a USB, you can see that it's in and unplug it before the robot turns on.

PR #271, opened 2022-11-23
Change from using Affine to Isometry in Eigen

TrentHouliston:
As per [NUbots#986](https://github.com/NUbots/NUbots/pull/986)


Bidski (approved):
LGTM

PR #274, opened 2023-02-20
Add instructions for NUpbr

ysims:
Up to date instructions. The readme instructions don't seem to work, but this seems to consistently work.


Crafty15 (approved):
Tested this personally. LGTM

BaileyJoe (approved):
Looks consistent with my process the last time I installed NUpbr. LGTM.

jpptm:
Do we want to account for non-members asking for the resources folder now? How do we want to possible deal with that in the future?

ysims:
> Do we want to account for non-members asking for the resources folder now? How do we want to possible deal with that in the future?

Not really, I think there would be copywrite issues with distributing the resources folder publicly. We could consider designing a new folder without these issues.

PR #275, opened 2023-03-05
Fix broken image in NUpbr guide

ysims:
Wasn't put in correctly. This should fix it.


jpptm (approved):
LGTM, change follows syntax in https://kabartolo.github.io/chicago-docs-demo/docs/mdx-guide/writing/

PR #281, opened 2023-08-31
Use archived iso matching dockerfile and update network interface names

JesseWilliamson:
Earlier this year I had a problem which I think was caused by disks being formatted with a newer version of `util-linux` than is provided by our Dockerfile and installed by systemconfigurator. To avoid this, it's probably best to stick with an iso from the same date as our mirrorlist.

Also, the network interfaces needed updating


Bidski:
With your changes in https://github.com/NUbots/NUbots/pull/1174 this should be irrelevant. I'm not sure this would actually do anything as I think the ISO uses a more standard mirrorlist without any date locking applied so it will likely always install the most up to date versions of things (unless you have no internet connection???)

JesseWilliamson:
> With your changes in [NUbots/NUbots#1174](https://github.com/NUbots/NUbots/pull/1174) this should be irrelevant. I'm not sure this would actually do anything as I think the ISO uses a more standard mirrorlist without any date locking applied so it will likely always install the most up to date versions of things (unless you have no internet connection???)

I think the problem was that when [mkfs is used in archinstall](https://github.com/NUbots/NUbots/blob/803610a88070b6f7eae9d32aa837346db122b5f2/doc/ArchInstall/arch_install.sh#L20), it is provided by whatever version of util-linux is already in the live install environment. This wouldn't be affected by the changes in NUbots/NUbots#1174

ysims (approved):

PR #288, opened 2023-09-30
Remove references to Darwin, CM-740

ysims:
This is an old branch that originally just removed the references to `darwin` when we removed it in the main codebase. It's been a while so I had to do some extra updates since integrating the OpenCR and phasing out the CM-740. One of the big aspects of this PR was making the NUClear examples more up to date so `darwin` wasn't mentioned and the links work. They are linked to a specific commit and are a little old but since they're just examples I think it's fine.

### Preview

<https://deploy-preview-288--nubook.netlify.com/system/foundations/nuclear> probably has the most changes but there's a few changed pages, see files changed for all of them.


Crafty15 (approved):
LGTM

PR #243, opened 2022-07-23
Update roles list after removal and moving of some roles

ysims:
For NUbots/NUbots#907
And NUbots/NUbots#866

PR #289, opened 2023-09-30
Update the Camera Calibration page after DataLogging change

ysims:
An old branch which is not as important after PR #243 but it still has some good updates with the update to DataLogging.

### Preview

<https://deploy-preview-289--nubook.netlify.com/guides/main/camera-calibration>


Crafty15 (approved):
LGTM

Crafty15 (approved):

PR #291, opened 2023-09-30
Update system diagram on overview page

ysims:
The system diagram is a fair bit out of date now that we have Director. I've updated the diagram but not sure if I like it. I wanted to have one diagram but nested clusters in GraphViz have limited functionality, for example arrows between clusters (as opposed to nodes) don't work nested. 

Up for any suggestions.

### Todo

- [ ] Make it nicer

### Preview

<https://deploy-preview-291--nubook.netlify.com/system/foundations/overview>


Crafty15 (approved):
LGTM

PR #297, opened 2023-10-07
Update algolia docsearch component to V3

LachlanCourt:
Several dramas with the existing docsearch component lead to me starting investigation around updating it a LONG time ago, I finally got around to finishing it today 🎉 
It looks a little bit different but I've tried to match the styles as closely as possible, this is totally open to feedback, I was trying to juggle minimal customisations vs looking nice so let me know if you think something looks off and if you have any suggestions :) 

### Preview
Note that the links that will show up in the search will link to the active site, not back to the preview so if you are clicking around just make sure you keep an eye on the URL.
Pls also consider checking on different screen sizes, mobile, tablet etc

https://deploy-preview-297--nubook.netlify.com/

Algolia


ysims:
Looks good on mobile. On desktop I noticed the colour of the background when the search is clicked is blue-ish. I feel like a grey would be better, particularly when in dark mode.

LachlanCourt:
What part of the search are you clicking sorry? I don't seem to be able to reproduce 🤔

src/components/header/search/search.jsx
```diff
@@ -1,33 +1,16 @@
...
-
-import { initialiseAlgolia } from './algolia-search'
-
-const Search = ({ background }) => {
-  useEffect(() => {
-    initialiseAlgolia()
-  }, [])
-
+import { DocSearch } from '@docsearch/react'
+const Search = () => {
```
JosephusPaye, line 6:
```suggestion
import { DocSearch } from '@docsearch/react'

const Search = () => {
```

src/components/header/search/search.jsx
```diff
@@ -1,33 +1,16 @@
...
-    <div
-      className={`${style.search} ${
-        background === 'solid'
-          ? style.backgroundSolid
-          : style.backgroundTransparent
-      }`}
-    >
-      <input
-        id='search'
+    <div className={`${style.search}`}>
```
JosephusPaye, line 8:
```suggestion
    <div className={style.search}>
```

src/components/header/search/search.jsx
```diff
@@ -1,33 +1,16 @@
...
+      <DocSearch
+        appId='BH4D9OD16A'
+        apiKey='be16e460d9d03fa711df82d525dec3c1'
+        indexName='nubots'
```
JosephusPaye, line 12: Perhaps something like this will fix the problem with search result URLs pointing to the main site (instead of localhost/deploy preview):

```jsx
<DocSearch
  // ... other props
  transformItems={(items) => {
    return items.map((item) => {
      // Get the origin (domain + possible port) from the result link
      const { origin } = new URL(item.url);
      return {
        ...item,
        // Replace the origin in the result URL with our current origin,
        // so the links work for local development and deploy previews
        url: item.url.replace(origin, window.location.origin),
      }
    });
  }}
/>
```

It's using the [`transformItems`](https://docsearch.algolia.com/docs/api#transformitems) prop to update the URLs on the search results before showing them.

src/components/header/search/search.module.css
JosephusPaye, line 1: Played around and made some minor adjustments to improve the styling (mostly tweaking colors):

<details><summary>Show adjusted styles</summary>

```css
/* Inline #0 | https://deploy-preview-297--nubook.netlify.app/ */

.DocSearch-Button {
  margin: 0;
}

.DocSearch-Button-Container {
  width: 100%;
}

.search-module--search--yrz9H > .DocSearch-Button > .DocSearch-Button-Container {
  /* gap: 1rem; */
  gap: 0.75rem;
}

@media (min-width: 768px) {
  .search-module--search--yrz9H > .DocSearch-Button {
    padding-left: 1rem;
  }
}

.dark .DocSearch-Modal {
  /* background-color: rgba(66,66,66,var(--tw-bg-opacity)); */
  background-color: #222;
}

.dark .DocSearch-Modal > .DocSearch-SearchBar > .DocSearch-Form {
  /* background-color: rgba(97,97,97,var(--tw-bg-opacity)); */
  /* --tw-bg-opacity: 1; */
  background-color: rgba(255,255,255,0.075);
}

.dark .DocSearch-Modal > .DocSearch-Dropdown > .DocSearch-Dropdown-Container > .DocSearch-Hits > .DocSearch-Hit-source {
  /* background-color: rgba(66,66,66,var(--tw-bg-opacity)); */
  background-color: #222;
}

.dark {
  /* --docsearch-hit-background: gray-800; */
  /* --docsearch-hit-active-color: rgba(50,50,50,.87); */
  /* --docsearch-highlight-color: #cc880d; */
  /* --docsearch-hit-shadow: 0 1px 3px 0 hsla(0,0%,100%,.45); */
  --docsearch-hit-background: rgba(255, 255, 255, 0.07);
  --docsearch-hit-active-color: #000;
  --docsearch-highlight-color: #f9a50d;
  --docsearch-hit-shadow: none;
}

.light {
  /* --docsearch-hit-active-color: rgba(50,50,50,.87); */
  --docsearch-hit-active-color: #fff;
}

.DocSearch-Hits mark {
  font-weight: 600;
}
```
</details>

Firefox's Devtools has a nice feature that you can use to preview those changes quickly:

Open the devtools, go to **Style Editor** tab, click `+` icon at the top of the left sidebar to add a new custom stylesheet, then copy and paste the code from above to apply the changes.

I made these adjustments manually in the browser, so I'll leave it up to you on where/how to add them to the NUbook stylesheet 😛

JosephusPaye (changes requested):
Switching to the Algolia v3 React component is a great idea - it's less code we have to maintain!

Your changes are also looking pretty good so far. I've just got a few styling suggestions and a possible fix for the results not linking to the current server (local or deploy previews).

ysims:
@LachlanCourt 

image

Outside of the search box, the rest of the site gets a blue-grey overlay.

src/components/header/search/search.jsx
LachlanCourt: That did the trick, nice find!

LachlanCourt:
Have updated to a light grey now instead, lemme know what you think @ysims

static/algolia-search-event-listener.js
LachlanCourt, line 1: This is now built in logic with the new component

src/components/seo.jsx
```diff
@@ -19,18 +19,8 @@ function SEO({ description, lang, meta, keywords, title }) {
...
-              },
-            ]}
-            script={[
-              {
-                defer: true,
-                src: withPrefix('algolia-search-event-listener.js'),
-              },
-            ]}
+            link={[]}
+            script={[]}
```
JosephusPaye, line 23: Can probably just remove these

```suggestion
```

src/components/header/search/search.module.css
```diff
@@ -1,143 +1,140 @@
...
-
-.searchIcon,
-.searchInput,
-.searchInput::placeholder {
-  transition: all 0.2s ease-in-out;
-}
-
-.backgroundTransparent {
-  .searchInput {
+  > :global.DocSearch-Button {
```
JosephusPaye, line 2: You might be able to remove `> ` from this and the other selectors in here so the styles are a bit less coupled to the exact HTML structure.

That'll make us a bit more resilient to changes in the Algolia component. For example, since `>` is the direct child selector, an update in the Algolia component that adds another element between `.search` and `.DocSearch-Button` will break our styles:

```html
<!-- Works -->
<div class="search">
   <button class="DocSearch-Button">Search</button>
</div>

<!-- Breaks -->
<div class="search">
   <div class="DocSearch-Button-Wrapper">
     <button class="DocSearch-Button">Search</button>
   </div>
</div>
```

A similar suggestion is to reduce the nesting to not rely on the HTML structure if we don't have to. For example, `.DocSearch-Button-Container` and `.DocSearch-Button-Placeholder` can be siblings at the same level instead of parent and child. Same with `.DocSearch-SearchBar` and `.DocSearch-Form`, etc.

Basically unnest as much as you can and remove `>` until stuff breaks :P

static/algolia-search-event-listener.js
JosephusPaye: Less code, nice!

ysims:
Looks good!

ysims (approved):
:rocket:

JosephusPaye (approved):
LGTM!

PR #310, opened 2024-08-15
Update windows instructions

yippyy:
changed to Ubuntu 22.04.3
added more alerts

### Preview

<https://deploy-preview-310--nubook.netlify.app/guides/main/getting-started>

### DIFF
--- a/src/book/03-guides/01-main-codebase/01-getting-started.mdx
+++ b/src/book/03-guides/01-main-codebase/01-getting-started.mdx
@@ -246,8 +246,6 @@ If your computer struggles to build and crashes, do './b build -jx', where x is 
 
 </Alert>
 
-
-
 <Alert type='info'>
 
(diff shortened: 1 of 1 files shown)


src/book/03-guides/01-main-codebase/01-getting-started.mdx
```diff
@@ -235,6 +242,12 @@ To build the code, open a terminal and change into the NUbots code directory. Th
 
 <Alert type='info'>
 
+If your computer struggles to build and crashes, do './b build -jx', where x is the number of cores you want to use. A good recommended amount of cores is five, if you still struggle, perhaps use one core.
```
ysims, line 245:
```suggestion
If your computer struggles to build and crashes, run `./b build -jx`, where x is the number of cores you want to use. A good recommended amount of cores is five, but if you still struggle perhaps use one core.
```

ysims (approved):

PR #325, opened 2025-04-27
Shuffle webots guide order

Tom0Brien:
This PR moves the `Running a NUbots Environment` section to top of tutorial.

This is the most common use case for `webots` and easier for people to follow when getting started, so better to put at the top.


ysims (approved):
Haha I did this today when updating the guide for the multi robots.  I'll be updating it a bit more to explain the NUWebots stuff, as it's not very clear by itself and most things are under the robocup env section.

Issue #284, opened 2023-09-27
Update input/cameras section with current settings


ysims:
The table is out of date 
https://nubook.nubots.net/system/subsystems/input

PR #336, opened 2025-09-04
Update the camera settings to reflect the current configuration files

ysims:
Addresses #284. Both the lens and general configs have been changed since. 

<https://deploy-preview-[PR_NUMBER]--nubook.netlify.app/system/subsystems/input>

### DIFF
--- a/src/book/02-system/02-subsystems/01-input.mdx
+++ b/src/book/02-system/02-subsystems/01-input.mdx
@@ -25,26 +25,33 @@ The cameras have the following parameters that are used by the object detection 
 | `lens.projection`   | string               | The lens projection type. Can be rectilinear, equidistant or equisold.                                                                                |
 | `lens.focal_length` | float                | The normalised focal length. It is defined as focal length in pixels divided by image width. The focal length is the angle of view and magnification. |
 | `lens.center`       | 2-dimensional vector | The normalised image centre offset. Represents the pixels from the centre of the image to the optical axis, divided by the image width                |
-| `k`                 | 2-dimensional vector | The polynomial distortion coefficients for the length                                                                                                 |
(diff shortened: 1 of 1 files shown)


JohanneMontano:
> > I thought roll and pitch offsets were an extrinsics thing related to the actual camera pose?
> 
> They are. It isn't the full pose transform though because the URDF is used to determine that, so any remaining offsets are just from however much it's off from the model.


```| `roll_offset`       | float (radians)      | Roll of the camera in radians to account for physical differences from manufactoring and assembly to align with the camera's virtual focal point.     |
| `pitch_offset`      | float (radians)      | Pitch of the camera in radians to account for physical differences from manufactoring and assembly to align with the camera's virtual focal point.  ```

Maybe just remove manufacturing and just do camera assembly? Manufacturing errors or discrepancies to me sounds like an intrinsics problem, which would relate to lens properties and lens mounting (even though we dont correct for skew)

ysims:
> > > I thought roll and pitch offsets were an extrinsics thing related to the actual camera pose?
> > 
> > 
> > They are. It isn't the full pose transform though because the URDF is used to determine that, so any remaining offsets are just from however much it's off from the model.
> 
> `` | `roll_offset` | float (radians) | Roll of the camera in radians to account for physical differences from manufactoring and assembly to align with the camera's virtual focal point. | | `pitch_offset` | float (radians) | Pitch of the camera in radians to account for physical differences from manufactoring and assembly to align with the camera's virtual focal point. ``
> 
> Maybe just remove manufacturing and just do camera assembly? Manufacturing errors or discrepancies to me sounds like an intrinsics problem, which would relate to lens properties and lens mounting (even though we dont correct for skew)

Ah, I mean NUgus manufacturing error. The 3D prints wont be perfect.

jpptm (approved):

PR #343, opened 2026-03-07
Update modules section with readmes generated from the main codebase

ysims:
Seems to have been a while since this was done.

<https://deploy-preview-343--nubook.netlify.app/system/modules/actuation/>

PR #345, opened 2026-03-17
Punch/fix latex formatting

miles-p:
On the Localisation module page, several strings of LaTeX don't appear correctly; appearing as text rather than formatted maths. This may be due to the fact that some Markdown/MDX viewers are more forgiving with simple LaTeX syntax errors.
This PR removes spaces between the `$$` tags and the actual LaTeX body, as well as using proper sub-script notation `u*{x}`.

### Preview

<https://deploy-preview-345--nubook.netlify.app/system/modules/localisation/>

### DIFF
--- a/src/book/02-system/05-modules/04-localisation.mdx
+++ b/src/book/02-system/05-modules/04-localisation.mdx
@@ -147,7 +147,7 @@ The optimization framework integrates several cost components and constraints to
 
    - Measures the alignment of predicted field lines with observed ones.
    - Calculated as the squared Euclidean distance between field line points and the nearest point on any observed line, scaled by a predefined weight:
-     $$ J_{\text{fl}} = w_{\text{fl}} \sum_{i=1}^{N} \left( \text{dist}(r_{\text{obs}_i}, r_{\text{line}}) \right)^2 $$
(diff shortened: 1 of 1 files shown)


ysims:
This should be fixed on the NUbots/NUbots side -- all of the module pages are automatically generated from the module readmes in the NUbots/NUbots repository using a script. Otherwise the issue will reappear the next time someone generates the pages.

ysims:
If you do put up a PR to fix it on the NUbots/NUbots side @miles-p, could you fix this one too while you're there? https://github.com/NUbots/NUbook/pull/343#discussion_r2937944576

miles-p:
I was unaware this was how NUbook entries are generated! Yes - I can fix it from there instead. Thanks Ysobel.
I will close this PR and take a look from the NUbots/NUbots side.

ysims:
@miles-p Only the ones in system/module are auto-generated. We had all the readmes already for each module in the codebase and decided it would be good to duplicate it on NUbook. 

The script still needs to manually run, which is what my PR #343 is. You can keep this PR and merge it, or I can update #343 once you get a PR into NUbots/NUbots.

miles-p:
This was solved by #343.
Thanks @ysims.

Issue #307, opened 2024-07-03
Allow creation of new section without updating js code for icons


ysims:
I think we should be able to create new sections just by adding the folders and mdx file. At the moment it breaks the site because it's expecting an icon to be specified for the sidebar in a js file. Maybe a default icon/no icon.

PR #352, opened 2026-05-24
Simplifies new section creation

ysims:
When we added the module readmes to NUbook, we initially made them in a new section. This threw up an error at first, because there was no associated icon. Adding icons is manually done per section in JS, essentially harcoding what sections exist.

This PR generalises section creation such that new sections can be made without touching any JS. The icons live inside the section folder, so adding one is simple for users. Additionally, a default icon exists in `components/sidebar` as a fallback to prevent errors.   

Addresses issue #307.

<https://deploy-preview-352--nubook.netlify.app/> (there isn't any noticeable change to the actual site)

Here is an example of adding a test section:

Just make a new folder and at least one page:

image

Then it'll show up with the default icon:

image
image

Adding an icon:

image


ZeoNyph (approved):
tested it out, works as expected; part of the long-term switch to the K1 is to archive NUgus documentation, so this should definitely make things easier

LGTM!

Issue #296, opened 2023-10-07
Bibtex not showing booktitle for proceedings unless page/volume number exists


ysims:
https://github.com/NUbots/NUbook/pull/295#discussion_r1349460759_

PR #362, opened 2026-08-02
Fix issue with references not showing booktitle if volume or pages weren't included

ysims:
Addresses Issue #296.

`inproceedings` is rendered either as a _published_ or _unpublished_ talk - which it is, is determined based on what fields are filled. When volume/pages/issue isn't filled, it assumes it's _unpublished_. Unpublished doesn't render `booktitle`, it uses the `event` key instead. However, our plain BibTeX set up apparently cannot actually use the event key, so the unpublished branch on the IEEE class is redundant at best and causes issues at worst. This PR removes the if logic around published/unpublished `inproceedings`, and just uses the normal published model.

An example reference is added to the kitchen sink. Before, the booktitle wouldn't display. Now, it does.

<https://deploy-preview-362--nubook.netlify.app/kitchen-sink/referencing>

### DIFF
--- a/build-plugins/gatsby-transformer-nubook-bibliography/ieee.csl
+++ b/build-plugins/gatsby-transformer-nubook-bibliography/ieee.csl
@@ -181,7 +181,7 @@
       <if type="paper-conference speech" match="any">
         <choose>
           <!-- Published Conference Paper -->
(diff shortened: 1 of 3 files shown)


miles-p (approved):
LGTM.

Image

